PR #13, opened 2020-04-17
Initial version of command-line support for test driver

Ndemco:
The commits in the PR add command-line support for the test driver. They also make a few related changes to the testing infrastructure (`tester.cpp`). 

The code to process cmd-line args and configure the tester does **not** test for errors. That is, the changes work only if the cmd-line options are specified the way they should be according to [requirements](https://github.com/sigcpp/stl-lite/wiki/Command-line-interface-for-the-test-driver). Error-handling and other special circumstances will be addressed in another PR.

Even though this PR merges relatively small number of changes from `add-cmd-line-nominal` branch to `add-cmd-line` branch, please review as if the changes in this PR and the changes in `add-cmd-line` are both being merged into `dev`. The reason for this approach is to keep `add-cmd-line` branch active for error-handling etc.

Changes to `driver.cpp` (renamed from `main.cpp`):
* Return the number of tests failed as the value of function `main`
* Process cmd-line arguments and configure the tester 

Changes to `tester.cpp`:

* Add function `getTestsFailed`
* Add function `setOutput`

### DIFF
--- a/test/driver.cpp
+++ b/test/driver.cpp
@@ -23,14 +23,18 @@
 
 #include "tester.h"
 
+using std::string;
+using std::string_view;
+using std::ios;
+
 //must be defined in a unit-specific source file such as "array-test.cpp"
 void runTests();
 
-void processCmdLine(std::vector<std::string_view> arguments);
+void processCmdLine(const std::vector<string_view>& arguments);
 
 int main(int argc, char* argv[])
 {
-    std::vector<std::string_view> arguments;
+    std::vector<string_view> arguments;
     for (int i = 0; i < argc; ++i)
         arguments.emplace_back(argv[i]);
     
@@ -39,20 +43,20 @@ int main(int argc, char* argv[])
    return getTestsFailed();
 }
 
-static const std::string defaultHeaderText("Running $exe");
+static const string defaultHeaderText("Running $exe");
 
-void processCmdLine(std::vector<std::string_view> arguments)
+void processCmdLine(const std::vector<string_view>& arguments)
 {
    bool printHeader{ true };
    bool printSummary{ true };
-   std::string headerText(defaultHeaderText);
-   std::string passReportMode("auto");
+   string headerText(defaultHeaderText);
+   string_view passReportMode("auto");
    unsigned short failThreshold(0);
-   std::string_view fileOpenMode;
-   std::string outputFilename;
+   string_view fileOpenMode;
+   string outputFilename;
 
    // begin parsing cmd-line arguments
-   for (std::size_t i = 1; i < arguments.size(); i += 2)
+   for (size_t i = 1; i < arguments.size(); i += 2)
    {
       if (arguments[i] == "-h")
          printHeader = arguments[i + 1] != "no";
@@ -64,7 +68,7 @@ void processCmdLine(std::vector<std::string_view> arguments)
          printSummary = arguments[i + 1] != "no";
       else if (arguments[i] == "-t")
       {
-         std::string_view& v = arguments[i + 1];
+         const string_view& v = arguments[i + 1];
          std::from_chars(v.data(), v.data() + v.size(), failThreshold);
          setFailThreshold(failThreshold);
       }
@@ -75,6 +79,7 @@ void processCmdLine(std::vector<std::string_view> arguments)
       }
    } //done parsing cmd-line arguments
 
+   //set pass report mode to the proper value
    if (passReportMode == "detail")
        setPassReportMode(passReportMode::detail);
    else if (passReportMode == "indicate")
@@ -83,6 +88,7 @@ void processCmdLine(std::vector<std::string_view> arguments)
        setPassReportMode(passReportMode::none);
    else if (passReportMode == "auto")
    {
+      //if output to file option enabled
       if (!fileOpenMode.empty())
          setPassReportMode(passReportMode::none);
       else
@@ -92,46 +98,53 @@ void processCmdLine(std::vector<std::string_view> arguments)
    //extract the exe name without path or filename extension
    //the result is used to expand the macro $exe
    std::filesystem::path exePath(arguments[0]);
-   std::string filenameNoExtension = exePath.replace_extension("").filename().string();
+   string filenameNoExt = exePath.replace_extension("").filename().string();
 
-   //replace $exe macro in header text
+   //replace $exe macro in header text only if a header text was defined
+   //and the print header option is enabled
    if (printHeader && !headerText.empty())
    {
-      std::string::size_type macroPosition = headerText.find("$exe");
-      for (; macroPosition != std::string::npos; macroPosition = headerText.find("$exe"))
-         headerText.replace(macroPosition, 4, filenameNoExtension);
+      string::size_type pos = headerText.find("$exe");
+      for (; pos != string::npos; pos = headerText.find("$exe"))
+         headerText.replace(pos, 4, filenameNoExt);
    }
+   //else do not print a header text
    else
       headerText = "";
 
    setHeaderText(headerText);
 
-   //replace $exe macro in filename
+   //if a filename was supplied, replace $exe macro in filename
    if (!outputFilename.empty())
    {
-      std::string::size_type macroPosition = outputFilename.find("$exe");
-      for (; macroPosition != std::string::npos; macroPosition = outputFilename.find("$exe"))
-         outputFilename.replace(macroPosition, 4, filenameNoExtension);
+      string::size_type pos = outputFilename.find("$exe");
+      for (; pos != string::npos; pos = outputFilename.find("$exe"))
+         outputFilename.replace(pos, 4, filenameNoExt);
    }
 
    // the file stream must be alive when runTests() is called
    std::ofstream fileOut;
+   //if output to file option not enabled, use standard output
    if (outputFilename.empty())
       setOutput(std::cout);
    else
    {
       setOutput(fileOut);
 
-      std::filesystem::path outputFilePath = outputFilename;
-      if (outputFilePath.extension() == "")
-         outputFilePath.replace_extension(".out");
-      if (fileOpenMode == "-f" && std::filesystem::exists(outputFilePath))
+      std::filesystem::path fileOutPath = outputFilename;
+      //if not extension supplied, append the default ".out" extension
+      if (fileOutPath.extension() == "")
+         fileOutPath.replace_extension(".out");
+      //if -f option enabled, make sure the file doesn't already exist
+      if (fileOpenMode == "-f" && std::filesystem::exists(fileOutPath))
       {
          std::cerr << "Output file alredy exists";
          return;
       }
 
-      fileOut.open(outputFilePath, fileOpenMode == "-fa" ? std::ios::app : std::ios::out);
+      //if -fa option is enabled, set open mode to append
+      fileOut.open(fileOutPath, fileOpenMode == "-fa" ? ios::app : ios::out);
+      //check for errors opening file
       if (!fileOut.is_open())
       {
          std::cerr << "Error opening output file";
@@ -148,6 +161,7 @@ void processCmdLine(std::vector<std::string_view> arguments)
       std::cout << msg << '\n'; //TODO: print in appropriate stream based in cmd-line
    }
 
+   //if print summary option is enabled
    if (printSummary)
       summarizeTests();
 }
--- a/test/driver.cpp
+++ b/test/driver.cpp
@@ -77,8 +77,7 @@ void processCmdLine(char* arguments[], const size_t size)
          printSummary = (value == option_value_yes);
       else if (option_name == option_name_threshold)
       {
-         const string_view& v = arguments[i + 1];
-         std::from_chars(v.data(), v.data() + v.size(), failThreshold);
+         std::from_chars(value.data(), value.data() + value.size(), failThreshold);
          setFailThreshold(failThreshold);
       }
       else if (option_name == option_name_file || option_name == option_name_file_overwrite ||
--- a/test/driver.cpp
+++ b/test/driver.cpp
@@ -30,7 +30,7 @@ using std::size_t;
 //must be defined in a unit-specific source file such as "array-test.cpp"
 void runTests();
 
-void processCmdLine(char* arguments[], const size_t size);
+void processCmdLine(char* arguments[], const std::size_t size);
 
 void replace_all(std::string& str, const std::string& substr, 
    const std::string& new_substr);
--- a/test/driver.cpp
+++ b/test/driver.cpp
@@ -25,26 +25,26 @@
 
 using std::string;
 using std::string_view;
+using std::size_t;
 
 //must be defined in a unit-specific source file such as "array-test.cpp"
 void runTests();
 
-void processCmdLine(const std::vector<string_view>& arguments);
+void processCmdLine(char* arguments[], const size_t size);
+
+void replace_all(string& str, const string_view substr, 
+   const string& toReplace);
 
 int main(int argc, char* argv[])
 {
-    std::vector<string_view> arguments;
-    for (int i = 0; i < argc; ++i)
-        arguments.emplace_back(argv[i]);
-    
-    processCmdLine(arguments);
+    processCmdLine(argv, argc);
 
    return getTestsFailed();
 }
 
 static const string defaultHeaderText("Running $exe");
 
-void processCmdLine(const std::vector<string_view>& arguments)
+void processCmdLine(char* arguments[], const size_t size)
 {
    bool printHeader{ true };
    bool printSummary{ true };
@@ -54,38 +54,50 @@ void processCmdLine(const std::vector<string_view>& arguments)
    string_view fileOpenMode;
    string outputFilename;
 
+   constexpr string_view name_header{ "-h" }, name_header_text{ "-ht" },
+      name_summary{ "-s" }, namePRM{ "-p" }, name_file{ "-f" },
+      name_file_overwrite{ "-fo" }, name_file_append{ "-fa" },
+      name_threshold{ "-t" };
+
+   constexpr string_view value_yes{ "yes" };
+   constexpr string_view value_detail{ "detail" },
+      value_indicate{ "indicate" }, value_none{ "none" }, value_auto{ "auto" };
+
    // begin parsing cmd-line arguments
-   for (size_t i = 1; i < arguments.size(); i += 2)
+   for (size_t i = 1; i < size; i += 2)
    {
-      if (arguments[i] == "-h")
-         printHeader = arguments[i + 1] != "no";
-      else if (arguments[i] == "-ht")
-         headerText = arguments[i + 1];
-      else if (arguments[i] == "-p")
-         passReportMode = arguments[i + 1];
-      else if (arguments[i] == "-s")
-         printSummary = arguments[i + 1] != "no";
-      else if (arguments[i] == "-t")
+      string_view name(arguments[i]), value(arguments[i + 1]);
+      
+      if (name == name_header)
+         printHeader = (value == value_yes);
+      else if (name == name_header_text)
+         headerText = value;
+      else if (name == namePRM)
+         passReportMode = value;
+      else if (name == name_summary)
+         printSummary = (value == value_yes);
+      else if (name == name_threshold)
       {
          const string_view& v = arguments[i + 1];
          std::from_chars(v.data(), v.data() + v.size(), failThreshold);
          setFailThreshold(failThreshold);
       }
-      else if (arguments[i] == "-f" || arguments[i] == "-fo" || arguments[i] == "-fa")
+      else if (name == name_file || name == name_file_overwrite ||
+         name == name_file_append)
       {
-         fileOpenMode = arguments[i];
-         outputFilename = arguments[i + 1];
+         fileOpenMode = name;
+         outputFilename = value;
       }
    } //done parsing cmd-line arguments
 
    //set pass report mode to the proper value
-   if (passReportMode == "detail")
+   if (passReportMode == value_detail)
        setPassReportMode(passReportMode::detail);
-   else if (passReportMode == "indicate")
+   else if (passReportMode == value_indicate)
        setPassReportMode(passReportMode::indicate);
-   else if (passReportMode == "none")
+   else if (passReportMode == value_none)
        setPassReportMode(passReportMode::none);
-   else if (passReportMode == "auto")
+   else if (passReportMode == value_auto)
    {
       //if output to file option enabled
       if (!fileOpenMode.empty())
@@ -101,11 +113,10 @@ void processCmdLine(const std::vector<string_view>& arguments)
 
    //replace $exe macro in header text only if a header text was defined
    //and the print header option is enabled
+   constexpr string_view exeMacro{ "$exe" };
    if (printHeader && !headerText.empty())
    {
-      string::size_type pos = headerText.find("$exe");
-      for (; pos != string::npos; pos = headerText.find("$exe"))
-         headerText.replace(pos, 4, filenameNoExt);
+      replace_all(headerText, exeMacro, filenameNoExt);
    }
    else
       headerText = "";
@@ -115,9 +126,7 @@ void processCmdLine(const std::vector<string_view>& arguments)
    //if a filename was supplied, replace $exe macro in filename
    if (!outputFilename.empty())
    {
-      string::size_type pos = outputFilename.find("$exe");
-      for (; pos != string::npos; pos = outputFilename.find("$exe"))
-         outputFilename.replace(pos, 4, filenameNoExt);
+      replace_all(outputFilename, exeMacro, filenameNoExt);
    }
 
    // the file stream must be alive when runTests() is called
@@ -134,14 +143,14 @@ void processCmdLine(const std::vector<string_view>& arguments)
       if (fileOutPath.extension() == "")
          fileOutPath.replace_extension(".out");
       //if -f option enabled, make sure the file doesn't already exist
-      if (fileOpenMode == "-f" && std::filesystem::exists(fileOutPath))
+      if (fileOpenMode == name_file && std::filesystem::exists(fileOutPath))
       {
          std::cerr << "Output file alredy exists";
          return;
       }
       //if -fa option is enabled, set open mode to append
       using std::ios;
-      fileOut.open(fileOutPath, fileOpenMode == "-fa" ? ios::app : ios::out);
+      fileOut.open(fileOutPath, fileOpenMode == name_file_append ? ios::app : ios::out);
       //check for errors opening file
       if (!fileOut.is_open())
       {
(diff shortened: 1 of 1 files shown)


smurthys:
Thanks @Ndemco for the contribution. I plan to review the changes tomorrow.

@sigcpp/stl-lite: Please use [this link](https://github.com/sigcpp/stl-lite/compare/sigcpp:dev...add-cmd-line-nominal) to compare the `add-cmd-line-nominal` branch with `dev`.

smurthys:
BTW, it seems there are two merge conflicts. I'll send a Teams message to @sigcpp/stl-lite about a potential meeting to go over resolving conflicts (because this is the first instance) and also answer questions about linking to code segments etc. in review comments.

test/driver.cpp
```diff
@@ -23,14 +23,18 @@
 
 #include "tester.h"
 
+using std::string;
+using std::string_view;
+using std::ios;
```
smurthys, line 28: Move using declarations to the narrowest possible scope

test/driver.cpp
```diff
@@ -92,46 +98,53 @@ void processCmdLine(std::vector<std::string_view> arguments)
...
    if (printHeader && !headerText.empty())
    {
-      std::string::size_type macroPosition = headerText.find("$exe");
-      for (; macroPosition != std::string::npos; macroPosition = headerText.find("$exe"))
-         headerText.replace(macroPosition, 4, filenameNoExtension);
+      string::size_type pos = headerText.find("$exe");
+      for (; pos != string::npos; pos = headerText.find("$exe"))
+         headerText.replace(pos, 4, filenameNoExt);
    }
+   //else do not print a header text
```
smurthys, line 111: This comment is unnecessary

test/driver.cpp
```diff
@@ -148,6 +161,7 @@ void processCmdLine(std::vector<std::string_view> arguments)
       std::cout << msg << '\n'; //TODO: print in appropriate stream based in cmd-line
```
smurthys, line 161: Message needs to be printed to std::cout or fileOut depending on tester configuration. Assuming fileOut is set to
```suggestion
      (outputFilename.empty() ? cout : fileOut )<< msg << '\n';
```

smurthys:
Thanks @Ndemco for the changes. I appreciate your effort and patience in both learning and contributing.

@sigcpp/stl-lite : I constructed three alternative approaches of a simplified version of the driver to see which approach produces better result. In each approach, `main` calls a function `process_args` to get back a struct with options extracted. Here is a brief description of the three approaches along with a Compiler Explorer link where you can examine the results. I like you to **review the alternatives and opine** on which version you think is better to use.

PS: I'll post a different comment about a re-organization of code to separate extracting info from cmd-line to actually applying the options to tester. @Ndemco has already made this observation in
a call I had with him a few days ago.

Approach 1: [https://godbolt.org/z/2vEYoR](https://godbolt.org/z/2vEYoR)
- Pass `argv` as is to `process_args`
- Construct `string_view` for current name and value
- Compare name and value with C-string literals: calls `string_view` ctor for each C-string literal

Approach 2: [https://godbolt.org/z/wDzGLb](https://godbolt.org/z/wDzGLb)
- Variation of Approach 1
- Pass `argv` as is to `process_args`
- Use `constexpr string_view` for option names and well-known values
- Construct `string_view` for current name and value
- Compare name and value with `constexpr string_view` already constructed

Approach 3: [https://godbolt.org/z/rSwD22](https://godbolt.org/z/rSwD22)
- Variation of Approach 2
- Collect `argv` in a vector of `string_view`
- Pass vector to `process_args`
- Use `constexpr string_view` for option names and well-known values
- Use `string_view&` for current name and value
- Compare name and value with `constexpr string_view` already constructed

Ndemco:
Thank you for the thoughtful reply @smurthys . After reviewing the examples, I think the second version achieves a happy medium of readability and efficiency and would be best to use. While the third option seems like the most "C++-style" way to process the command line, I don't think the added overhead of including a vector is worth the benefits. Another option would be using a vector of c-strings, but there is nothing really gained in terms of efficiency as opposed to using a vector of `string_view` since the overhead comes from using a vector, not the `string_view` in the vector. Using a vector of c-strings would also take away the benefit of being able to use `string_view&` for the current name and value.

smurthys:
Thanks @Ndemco for reviewing the alternatives. 

I suggest function `processCmdLine` be revised along the lines of "Approach 2", which is to say pass `argv` and `argc` to `processCmdLine` as in "Approach 2". and revise `processCmdLine` minimally to:
- Create and use `constexpr string_view` objects for option names and well-known option values
- Create temporary `string_view` objects for option name and value being examined in the first `for` loop

Unlike "Approach 2", I recommend that we **not** revise `processCmdLine` to return an "options" structure. Instead, at this stage, let `processCmdLine` continue to call various `set` functions. We can incrementally make that change after the changes I propose here are made. (We will make the next incremental change in this PR, just not right away.)

Please let me know if you have questions.

test/driver.cpp
```diff
@@ -92,46 +97,52 @@ void processCmdLine(std::vector<std::string_view> arguments)
...
-   //replace $exe macro in filename
+   //if a filename was supplied, replace $exe macro in filename
    if (!outputFilename.empty())
    {
-      std::string::size_type macroPosition = outputFilename.find("$exe");
-      for (; macroPosition != std::string::npos; macroPosition = outputFilename.find("$exe"))
-         outputFilename.replace(macroPosition, 4, filenameNoExtension);
+      string::size_type pos = outputFilename.find("$exe");
+      for (; pos != string::npos; pos = outputFilename.find("$exe"))
+         outputFilename.replace(pos, 4, filenameNoExt);
```
smurthys, line 120: The code to expand the `$exe` macro occurs twice. Following the DRY principle (Don't Repeat Yourself), I recommend refactoring the repeated code to a function with a declaration such as the following. A careful review of the `std::string` docs could also help, but I don't have much insight to add on that at this very moment.

 `void replace_all(std::string& str, const std::string& substr);`

test/driver.cpp
```diff
@@ -83,6 +87,7 @@ void processCmdLine(std::vector<std::string_view> arguments)
        setPassReportMode(passReportMode::none);
    else if (passReportMode == "auto")
```
smurthys, line 88: Consider the following revisions:
- change `passReportMode` variable's type to the enum type of that name
- Refactor the code in L81-L87 to a new function `getPassReportMode` that accepts a `string_view` and returns `passReportMode` enum
- Use function `getPassReportMode` in L65

test/driver.cpp
Ndemco: Thank you for the feedback @smurthys, I agree that the repeated code should be refactored. Is there any reason the `substr` parameter can't be a `string_view` since it will only be read?

smurthys: Yes @Ndemco, the second parameter could be `string_view`. Also, I notice the declaration I provided is missing a parameter. I do advise writing the function in isolation and verify that it works before changing the larger code:

`void replace_all(std::string& str, const std::string_view& substr, const std::string_view replacement);`

Ndemco: @smurthys I've made the changes and pushed to origin. For `replace_all()` I used a `const string&` as the `toReplace` parameter since the text we want to replace the substring with (`filenameNoExt`) already exists as a string. I did try to change `filenameNoExt` to a `string_view` and then change the `toReplace` parameter to a `const string_view&` to match, but it caused bugs. For some reason when `filenameNoExt` is a `string_view`, its value is missing the first letter of the expected text; "test-stl-lite" becomes "est-stl-lite". For now I have left the variable as a `string` and the parameter as a `const string&` until I can investigate further.

test/driver.cpp
Ndemco: @smurthys I'm not sure how we would resolve when "auto" is supplied as the pass report mode as this requires knowing which file open mode the user has supplied as well which we have no way of knowing for sure until after all the command-line arguments have been parsed.

Perhaps I misread your suggestions but if not an alternative could be to keep `passReportMode` as a `string_view`, refactor the code in L93-L107 into a function `getPassReportMode` that takes in a reference to `passReportMode` and a reference to `fileOpenMode` and returns a `passReportMode` enum. Then use the return value of `getPassReportMode` as the parameter in `setPassReportMode`. I don't see much benefit to this, however, as this is essentially just taking a block of a code and moving it to a function that will only be used once. Perhaps it would increase readability.

Ndemco:
Here are some possible improvements I suggest for the code:

* The `substr` parameter for function `replace_all` on L177 can be a `const string_view&` rather than a `const string_view` to prevent creating a copy of the argument.

* Rather than `headerText` on L51 being initialized with `defaultHeaderText` (global variable on L45), it could just be initialized to "Running $exe" the same way the other variables are initialized to their defaults to avoid having an unnecessary global variable.

* The string_view variables declared starting on L57 all begin with "name"; the use of "name" could sometimes make what the variable represents unclear, for example, `name_file` when read can be mistaken for the name of a file rather than the "-f" file command. I propose the "name" part be changed to something like "cmd" to improve readability.

test/driver.cpp
```diff
@@ -163,3 +172,12 @@ void processCmdLine(const std::vector<string_view>& arguments)
...
 }
+
+//Replace all instances of a substring with a different substring
+void replace_all(string& str, const string_view substr, 
+   const string& toReplace)
+{
+   string::size_type pos = str.find(substr);
+   for (; pos != string::npos; pos = str.find(substr))
+      str.replace(pos, 4, toReplace);
+}
```
smurthys: Thanks for writing the new function. I agree the last parameter needs to be received by reference.

I see the following issues:
- The function needs to work for any substr, new substr values: it currently does not due to substr length being hardcoded
- The function should work in any context, not just for the limited context of our code. Thus, it is better to be uniform about using either string_view or string for the substr parameters.
- I have some guidelines for when to use const string& parameter vs const string_view&. I am yet to publish that post, but in this case const string& would work just fine. (A function template is actually ideal)
- L181: `pos = str.find(substr)` unnecessarily searches from the beginning of the string. The search can safely start from pos+substr_size, but that would ignore the possibility the new substr can be shorter. And, to ensure termination, we can reasonably assume that the new substr does not contain the replaced substr. We should add an assertion to that effect.

I have the following suggested rewrite of the function. An implementation with unit tests is here: https://godbolt.org/z/7ipMLX. 

Please verify my implementation and tests and let me know if there are any issues. Note that I have intentionally commented out the last test case which causes assertion failure.

```suggestion
//replace all instances of a substring with a different substring
//the new substring should not contain the replaced substring
void replace_all(std::string& str, const std::string& substr, const std::string& new_substr)
{
   assert(new_substr.find(substr) == std::string::npos);

   auto substr_size = substr.size(), new_substr_size = new_substr.size();
   auto pos = str.find(substr);
   for (; pos != std::string::npos; pos = str.find(substr, pos+new_substr_size))
      str.replace(pos, substr_size, new_substr);
}
```

smurthys:
Thanks @Ndemco for the changes. I will do another round of review later tonight. In the meantime, please check my proposed changes for the `replace_all` function.

@sigcpp/stl-lite: Going forward, whenever we create a function as part of PR, I recommend we write the function as standalone, write tests for the function, and share a CE link before pushing the changes. An example of this approach is my [recent comment](https://github.com/sigcpp/stl-lite/pull/13/commits/ed108044192f4236dd16185ab36da7e4e91fa2bc#r429584932). 

I realize this approach is not always practical, but let us endeavor to practice it whenever it is feasible.

test/driver.cpp
```diff
@@ -23,66 +23,83 @@
...
+using std::size_t;
+
 //must be defined in a unit-specific source file such as "array-test.cpp"
 void runTests();
 
-void processCmdLine(std::vector<std::string_view> arguments);
+void processCmdLine(char* arguments[], const size_t size);
+
+void replace_all(string& str, const string_view substr, 
+   const string& toReplace);
```
smurthys, line 36: I propose the following function declaration to go with the suggested change to the function's definition:

```suggestion
void replace_all(std::string& str, const std::string& substr, const std::string& new_substr);
```

smurthys:
@Ndemco I agree with the following changes you propose:

- Initialize variable `headerText` directly with a literal and remove the variable `defaultHeaderText`

- I propose the prefix `option_name_` for variables that hold option names and the prefix `option_value_` for variables that hold well-known option values. The prefix `cmd` is **not** recommended because the names and values relate to options on the cmd-line. The term "option" is consistent with the documentation for this feature.

test/driver.cpp
smurthys: @Ndemco Maybe I'm missing something but I don't see any issue with using the suggested function on L65 as I propose. Any prm the uses explicitly proposes is used regardless of the what that `-f*` option says. `auto` is the default prm no matter if/what `-f*` option is used. That is, using the value determined on L65 in place of the current code in L81-L87 shouldn't be a problem. The logic starting from L88 about prm being `auto` can still remain except that check will start the `if else` flow.

Having said this, I worry I'm missing something fundamental. I can try making the change if you like and perhaps that would help me see the error in my logic.

test/driver.cpp
smurthys: I thought a bit more about `replace_all` and concluded that the assertion is unnecessary. I also concluded that the offset used in repeated find could be substr_size in some cases, and new_substr_size in other cases. Also, there is the case where offsetting by substr_size can cause non-termination. (Interestingly, boost library's `replace_all` appears to offset by substr_size).

With these factors in mind, and with some optimization, I revised the implementation and the unit tests: https://godbolt.org/z/4VkDvU

**Please review the revised version and opine.**

smurthys:
BTW, we now have mixture of difference conventions for identifiers and possibly even for code layout. I propose we all review the sections of C++ Core Guidelines I listed in a SIGCPP Teams message and finalize the style.

test/driver.cpp
smurthys: I don't mean to spam, but I realized that the `replace_all` function was missing a choice of offset to use for "next find". 
For comparison, I also wrote a program using `boost::replace_all` to verify that the results are "correct". Here is the CE link: https://godbolt.org/z/cAM8i8

At this point, I **like some help testing the function** and comparing the results with boost. This mostlu entails constructing test cases. Note that the `replace_all` I wrote lets the programmer choose how and how much to skip over at each find operation. Thus, the result isn't incorrect if it differs from boost, but the difference only means the value passed to parameter `offset_by` isn't the same as what boost implicitly uses. In other words, the test cases should be such that they reveal which value of `offset_by` matches boost, and which value causes deviation from boost. **The outcome of the study will be choose the most appropriate default value for parameter `offset_by`.**

**PS:** It can also be helpful to compare the results with Java's `replace` method for strings: https://docs.oracle.com/en/java/javase/12/docs/api/java.base/java/lang/String.html#replace(java.lang.CharSequence,java.lang.CharSequence)

Ndemco: @smurthys I agree the next `substr` find should not start at the beginning of the string every time--not just for efficiency's sake--but if `substr` is a substring of  `new_substr`, the function would be caught in an endless loop.

For example, replacing all occurrences of "John" with "Johnathan". If the find begins at the beginning of the string every time, it would find the "John" in "Johnathan" and would change it to "Johnathanathan", with "athan" endlessly being added.

I also agree that the function should work in any context.

I have reviewed your revised implementation and believe that this is a much healthier version of `replace_all`, however, I'm having trouble envisioning a scenario in which a user would want to offset by `substr.size()` rather than `new_substr.size()`. Would you not always want to offset by the size of the new substring in order to avoid altering the text in the new substring?

I will work on creating a comprehensive test driver for `replace_all` that compares the results to `boost::replace_all` and plan to have it done by this Friday.

Thank you for your suggestions and input.

test/driver.cpp
Ndemco: @smruthys I have made the changes and pushed to origin. Thanks for the feedback.

Ndemco:
@smurthys I've pushed some changes reflected in your suggestions [here](https://github.com/sigcpp/stl-lite/pull/13#issuecomment-633169146) that make some of the variable names more readable and resolve our concern with the use of a global variable.

test/driver.cpp
```diff
@@ -52,50 +52,51 @@ void processCmdLine(char* arguments[], const size_t size)
...
+   constexpr string_view option_value_yes{ "yes" };
+   constexpr string_view option_value_detail{ "detail" },
+      option_value_indicate{ "indicate" }, option_value_none{ "none" },
+      option_value_auto{ "auto" };
 
    //begin parsing cmd-line arguments
    for (size_t i = 1; i < size; i += 2)
    {
-      string_view name(arguments[i]), value(arguments[i + 1]);
+      string_view option_name(arguments[i]), value(arguments[i + 1]);
```
smurthys, line 68: I suggest reverting the identifier `option_name` back to just `name` for both simplicity and to be consistent with the name `value`. These two variables have pretty short scope and thus are quite visible.

test/driver.cpp
```diff
@@ -23,66 +23,82 @@
...
+using std::size_t;
+
 //must be defined in a unit-specific source file such as "array-test.cpp"
 void runTests();
 
-void processCmdLine(std::vector<std::string_view> arguments);
+void processCmdLine(char* arguments[], const size_t size);
+
+void replace_all(string& str, const string& substr, 
+   const string& toReplace);
```
smurthys, line 36: I suggest changing the parameter names in this function declaration to match the names used in the function's definition.

```suggestion
void replace_all(string& str, const string& substr, 
   const string& new_substr);
```

test/driver.cpp
```diff
@@ -23,66 +23,82 @@
...
+         headerText = value;
+      else if (option_name == option_name_prm)
+         passReportMode = value;
+      else if (option_name == option_name_summary)
+         printSummary = (value == option_value_yes);
+      else if (option_name == option_name_threshold)
       {
-         std::string_view& v = arguments[i + 1];
+         const string_view& v = arguments[i + 1];
          std::from_chars(v.data(), v.data() + v.size(), failThreshold);
```
smurthys, line 81: I believe there is no longer a need to use the temporary variable `v`:

```suggestion
         std::from_chars(value.data(), value.data() + value.size(), failThreshold);
```

test/driver.cpp
```diff
@@ -23,66 +23,82 @@
 
 #include "tester.h"
 
+using std::string;
+using std::string_view;
+using std::size_t;
```
smurthys, line 28: I recommend moving these declaration to function `processCmdLine`.

test/driver.cpp
```diff
@@ -145,9 +164,19 @@ void processCmdLine(std::vector<std::string_view> arguments)
...
       summarizeTests();
+}
+
+//Replace all instances of a substring with a different substring
+void replace_all(string& str, const string& substr, 
+   const string& new_substr)
+{
+   string::size_type pos = str.find(substr);
+   for (; pos != string::npos; pos = str.find(substr, pos + substr.size()))
+      str.replace(pos, 4, new_substr);
```
smurthys, line 181: I believe this function definition will change based on my earlier observations about generalizing it, but it still needs to be correct in its current state: the last line of the function cannot use the fixed size `4`.

```suggestion
void replace_all(std::string& str, std::const string& substr, 
   const std::string& new_substr)
{
  auto pos = str.find(substr);
  auto substr_size = substr.size();
   for (; pos != std::string::npos; pos = str.find(substr, pos + substr_size))
      str.replace(pos, substr_size, new_substr);
```

test/driver.cpp
```diff
@@ -23,66 +23,82 @@
...
 
+using std::string;
+using std::string_view;
+using std::size_t;
+
 //must be defined in a unit-specific source file such as "array-test.cpp"
 void runTests();
 
-void processCmdLine(std::vector<std::string_view> arguments);
+void processCmdLine(char* arguments[], const size_t size);
```
smurthys, line 33:
```suggestion
void processCmdLine(char* arguments[], const std::size_t size);
```

test/driver.cpp
```diff
@@ -23,66 +23,82 @@
...
-    processCmdLine(arguments);
+    processCmdLine(argv, argc);
 
    return getTestsFailed();
 }
 
-static const std::string defaultHeaderText("Running $exe");
-
-void processCmdLine(std::vector<std::string_view> arguments)
+void processCmdLine(char* arguments[], const size_t size)
```
smurthys, line 45:
```suggestion
void processCmdLine(char* arguments[], const std::size_t size)
```

smurthys:
Thanks again @Ndemco for the revisions. Please consider the following suggestions

- Review the suggested revisions I made today and accept them where you agree. Please feel free to opine or ask questions if you disagree or need clarification
- Remove any unnecessary header file inclusion
- Consider implementing a function related to `passReportMode` that is discussed in an [earlier thread](https://github.com/sigcpp/stl-lite/pull/13#discussion_r423810306)
- Avoid creating blocks with just one statement in them, For example: [L127-129](https://github.com/sigcpp/stl-lite/blob/cae625fd5b0f6a41d39ebf538ca35d81bc97e854/test/driver.cpp#L127-L129)
- Thoroughly review code: as I have mentioned a few times before, it is a good idea to review code thoroughly especially as we fix issues.

There are likely other issues to address, but I recommend addressing the issues identified thus far before finding/fixing new ones.

Also, please make a practical timeline for fixing the issues identified so far so that we can plan the remaining work.

Thanks again for your contribution.

test/driver.cpp
Ndemco: Agreed! Oversight on my part.

test/driver.cpp
Ndemco: You are correct, there is no longer a need for variable `v`.

Ndemco:
@smurthys I will first focus on resolving the issues brought up [here](https://github.com/sigcpp/stl-lite/pull/13#issuecomment-634983450), then the test driver for the new `replace_all` function mentioned [here](https://github.com/sigcpp/stl-lite/pull/13#discussion_r430011114).

Let me know if you think I should focus on testing the `replace_all` function first. Either way, I will get both of these tasks done by the end of the week.

Thanks for all of your input.

smurthys:
@Ndemco We can study and improve `replace_all` after addressing other issues. Until then let us make only the changes necessary to that function.

test/driver.cpp
smurthys: The choice of the offset depends on the app. I agree that offset by `new_substr` size is a good choice. Please make that change.

test/driver.cpp
Ndemco: @smurthys There is no `passReportMode::auto`, so in the `getPassReportMode` function you proposed, if the `string_view` parameter passed to the `getPassReportMode` function is `"auto"`, in order for the function to return a `passReportMode` enum, it would have to know whether the user has chosen to send the results to a file or not by checking if `fileOpenMode` is empty. At first I thought this could be solved by including a reference to `fileOpenMode` in the function parameters, but after thinking on it, I realized that the wrong value for `fileOpenMode` will be passed to `getPassReportMode` if the `-f` option is read after the `-p` option.

Another issue I see is that because there is no `passReportMode::auto`, if the variable `passReportMode`  was set to the enum type of that name, rather than a `string_view` as it is now, we would not be able to give it a default value of `auto`.

It is possible I'm still misunderstanding; if you believe so, it might be best for you to try to make the change so both of us can see where the misunderstanding is.

smurthys: Thanks @Ndemco for helping me see the issue. I had completely missed that `auto` is not one of the enumerated values, and we should not add that as a value. 

I still wrote the function `getPassReportMode` to convert text prm to enum, and used the new function. Please review my changes. They are in commit adca120

Ndemco: @smurthys  The changes look great! Can the parameter `const std::string_view value` not be a reference to a `string_view`? (ie. `const std::string_view& value`)

smurthys: Thanks @Ndemco for spotting the omission. I just pushed the changes to address the issue. Please review.

smurthys: I reviewed the code around setting prm and changed fn. `getPassReportMode` to handle auto based on whether or not the output goes to a file. This cleans up the code in `processCmdLine` and enables us to refactor `processCmdLine`. Also in `processCmdLine`, the prm variable is default initialized so accurately reflect the state. `getPassReportMode` interprets empty prm value as "auto" consistent to the documentation.

test/driver.cpp
```diff
@@ -145,9 +146,46 @@ void processCmdLine(std::vector<std::string_view> arguments)
...
    }
 
+   //if print summary option is enabled
    if (printSummary)
       summarizeTests();
-}
\ No newline at end of file
+}
+
+//Replace all instances of a substring with a different substring
```
smurthys: Comment does not need to begin with upper case letter: no need to treat comments as sentences

```suggestion
//replace all instances of a substring with a different substring
```

test/driver.cpp
smurthys: I just made the change to offset next find by `new_substr` size.

Ndemco: Looks good!

test/driver.cpp
Ndemco: Looks good. Should we initialize `string_view passReportMode` to `"auto"` rather than keeping it empty? I understand you've written `replace_all` to treat an empty string as `auto`, but perhaps initializing `passReportMode` to `"auto"` makes it more explicit that the `auto` option is the default option. We could still leave the code that treats an empty `string_view` as `auto` for the sake of defensive programming.

smurthys: I default-initialized the prm variable intentionally, but I see your point. Please make a suggestion in a comment and I will accept it.

PS: I assume you meant function `getPassReportMode` instead of `replace_all` as treating empty prm value as "auto".

Ndemco: Yes, my mistake! I meant `getPassReportMode`.

test/driver.cpp
```diff
@@ -15,123 +15,124 @@
...
+   using std::string_view;
+   using std::size_t;
+   
+   //default values for cmd-line option names in name-value pairs
    bool printHeader{ true };
    bool printSummary{ true };
-   std::string headerText(defaultHeaderText);
-   std::string passReportMode("auto");
+   string headerText("Running $exe");
+   string_view passReportMode;
```
Ndemco, line 51: As discussed [here](https://github.com/sigcpp/stl-lite/pull/13#discussion_r434258887), I suggest we initialize `passReportMode` to `"auto"` to make it more explicit that `auto` is the default pass report mode.
```suggestion
   string_view passReportMode("auto");
```

smurthys: Thanks @Ndemco for the suggestion. As I said earlier, I see your point, but I'm a bit worried that the literal "auto" now appears in two different places, far from each other. The way it is right now, only function `getPassReportMode` determines the exact spelling and logic of the default behavior. Implementing the suggestion changes that and introduces a point of failure in that if the default behavior and the text representing that is changed later, it needs to changed in two places and it is easy to forget changing one of them. What do you think?

Ndemco: Yes, I see what you're saying @smurthys. I agree there should only be one place that the code needs to be changed if the default behavior / text is changed. Perhaps we keep `passReportMode` as an empty `string_view` and simply add a comment the states function `getPassReportMode` treats an empty `string_view` as the default `auto` value.

PR #14, opened 2020-06-07
Modularize test driver

smurthys:
The changes in this PR "modularize" the test driver in the following ways:

- Use an `Options` structure
- Separate the code to extract options from the code to apply
- Dedicated source and header files for options-management code
- Simpler `driver.cpp` in preparation for exception management

Tester changes:
- New functions `log` and `logLine`

Tidy up all files touched:
- 110 max line length
- Tab width: 4 spaces
- Keep tabs
- K&R layout
- snake_case for identifiers (except in "tester")

Many thanks to @Ndemco and @cmcmone for their contributions.


smurthys:
@Ndemco and @cmcmone, since the last push by one of you:
- Renamed `Options.output_filename` and changed its type to `std::filesystem::path`
- Moved the code to set default file extension from `apply_options` to `get_options` 
- Edited a few comments and renamed a variable
- Renamed function `get_yes_no` to `strtobool` (no snake_case on purpose to match `strtol` etc.)

I tested the changes and everything seems to work fine. However, please review the code again and opine.

cmcmone:
I tested and everything is good, but I have a question,  `options.cpp` file included `options.h`, so can we delete the include file like `iostream`, `filesystem`, `string`, `string_view` because these include file already included in `options.h` and 'tester.h'(others such as `driver.cpp` have same issue).

smurthys:
Quickly, the iostream inclusion in `options.h` seems extra. That needs to be removed.

No file should depend on includes in other files. Each file should explicitly include what it needs. Include guards will prevent the same file from being included more than once within a translation unit.

smurthys:
If you push changes to any PR, please add a comment drawing attention to the push and include a summary of changes. Thank you.

smurthys:
I just pushed changes that remove unnecessary includes from `options.h` and `tester.h`

cmcmone:
Ok, I will add a comment next time.
1. how about `string` and `string_view`. If `string` is included, do we still need to include ‘string_view’?
2. [`options.cpp` line 41-42 `std::` is unnecessary, because it's already declared in line 30-31](https://github.com/sigcpp/stl-lite/pull/14/commits/d9213ce8bc8f969cbfbafe605722015e14ced4d1#diff-5126592990b7dd1a17ac6c8c6454f157R30-R42).

smurthys:
Thanks for the pointer @cmcmone. I'd rather remove the using declarations and qualify the typenames with the namespace.

smurthys:
I just removed the `using` declarations in `get_options`.

Ndemco:
One possible improvement that would hold true to our M2 goal of designing such that exception handling is easily added later, is changing the code on L26 of `driver.cpp` such that `Options options` is not called in the same statement that its proper values are set in `get_options`. If our plan is to stick each function in its own try-catch block, `options` would be out of scope for `apply_options` which uses `options` as a parameter, and the conditional `if` statement before calling `summarizeTests`. This may be something we'd rather address in M3, or this may be a non-issue if we plan on formatting our exception handling differently.

Other than that, everything looks great! Thanks @smurthys and @cmcmone for your work on this milestone.

smurthys:
Good observation @Ndemco about the scope of `options` variable. However, there are a couple of engineering issues to consider there, and I think the choice should be made with the full picture of exception handling in M3, One consideration is RVO ("return value optimization", about which I have a draft blog post). 

Reviewers: Please approve the PR if you feel it is ready to merge. Of course, you are welcome to make additional observations, but I ask we do so soon. (M2 deadline was yesterday.)

smurthys:
@cmcmone Indeed, the <string> header includes <string_view> header, but there is no need to optimize so much because include guards will only include a header once. It is a lot better to be disciplined about including headers for each feature we use: keeps it simple and makes our code independent of third-party file organization.

PR #18, opened 2020-06-13
Support multiple test suites and add helpful verifiers

smurthys:
This PR adds support for multiple test suites (goal of M4) and also adds helpful verifiers. I welcome your review. Here is a summary of changes:

**Support for multiple test suites**:
- Introduce notion of "suite runner" which is any function with the declaration matching `void f()`, where `f` would be the name of a suite runner. `array_test` in `array-test.cpp` is an actual suite runner.
-  Function `run_suites` in `driver.cpp` sets up the tester for each suite and invokes test suites. See the comment associated with the `std::map` variable `suites` for a brief on how to specify the available "suite runners". Also see the comment just preceding the function about declaring suite runners. The specification of suite runners could possibly be better, but I feel the current state suffices for this PR.
- Add `-run` cmd-line option to specify the subset of test suites to run. The value of this option is a semi-colon separated list of suite names to run. This option will be quite helpful during the development of a test suite.
- Introduce the macro `$suite` to stand for suite name. At present the macro expansion is done only in header text. The next round of changes might support expanding this macro also in the file
- Change default header text to `Running $suite` from `Running $cmd` 
- Add new files `utils.h` and `utils.cpp` to hold truly general-purpose functions such as `replace_all` and the newly added function template `split`. The function template and its unit tests are at https://godbolt.org/z/LfPFVH. (EDIT: updated link with additional test cases)
- **Documentation changes are pending**. I will quite likely address them independent of this PR due to time shortage (and at least for now that change is in a separate repo anyway).

**Changes to tester:**
- Track suite-level tests-done and tests-failed
- Fail threshold is applied to total failed tests across all suites.
- Print suite-level summary
- Snake_case identifiers
- We need to discuss what if any cmd-line options need to be added

**New helpful verifiers:**
- Added: `is_true`, `is_false`, `is_zero`, `is_nonzero`, `is_negative`, `is_nonnegative`, `is_positive`, 
- Additional helpers are possible and we can add them as we recognize the need.
- Replaced calls to `verify` in array_test with is_true only for illustration. Using the helpers in that test suite is not part of this PR.
- Function `verify` should not be called directly, but that call is not prevented: nothing really wrong in calling it, but using the helpers makes tests simpler and somewhat self-documenting.


smurthys:
My apologies for the delay in creating this PR. I was first delayed by a meeting that went long. I was then sucked in by an "unhandled exception" and an issue with the new function template `split`. The fixes turned out to be fairly simple in both cases, but I did lose quite a bit of time locating the cause. I also spent a few minutes consoling myself that it was OK to not be able to see the obvious problem that cost me precious time. (What else could I do at this point?)

cmcmone:
Everything looks great! here is my question:
1. [line 89](https://github.com/sigcpp/stl-lite/pull/18/commits/aad13fc292289c21e7cb787fa8448ab9ffc03721#diff-041657b7c7f23065225f05f4f748fdf5R89) of `driver.cpp`, is it better to use std::unordered_map? Because the order of the keys does not seem to matter. 
2. [line 97](https://github.com/sigcpp/stl-lite/pull/18/commits/aad13fc292289c21e7cb787fa8448ab9ffc03721#diff-041657b7c7f23065225f05f4f748fdf5R97) of `driver.cpp`, do we need to determine whether options.suites_to_run is empty before calling the split function? if it's empty, we don't need call split function.
3. Why the'split' function is implemented in the header file and not in the '.cpp' file?
4. Some places did not use `UIS`. (Do we need to use `UIS` all the time?).
5. `test.cpp` [line 85](https://github.com/sigcpp/stl-lite/pull/18/commits/719b3fb18bfd9a37daee36a5f0dce211854408d0#diff-5e2c30b9e905f37296cccd6280d8fc73R85), do we need to throw an exception if the `name` is empty.

I will continue to review this PR tomorrow.

smurthys:
Thanks @cmcmone for the review:
> 1. [line 89](https://github.com/sigcpp/stl-lite/pull/18/commits/aad13fc292289c21e7cb787fa8448ab9ffc03721#diff-041657b7c7f23065225f05f4f748fdf5R89) of `driver.cpp`, is it better to use std::unordered_map? Because the order of the keys does not seem to matter.

Good point. I am not sure why I used map instead of unordered map. I will weigh the pros and cons.


> 2. [line 97](https://github.com/sigcpp/stl-lite/pull/18/commits/aad13fc292289c21e7cb787fa8448ab9ffc03721#diff-041657b7c7f23065225f05f4f748fdf5R97) of `driver.cpp`, do we need to determine whether options.suites_to_run is empty before calling the split function? if it's empty, we don't need call split function.

I thought about this quite a bit when implementing it, but decided not to test because there are many conditions under which a non-empty input can still result in an empty list of suites to run and it is not clear if "run all suites" should be true only if the input string is empty. What if the string is `;;` or just spaces? The better thing to do is to return early in `split` if the input is empty.


> 3. Why the'split' function is implemented in the header file and not in the '.cpp' file?

`split` is a function template.


> 4. Some places did not use `UIS`. (Do we need to use `UIS` all the time?).

Quite possible: old habit 😢. Also, initial values that from long/complex expressions are traditionally written with assignment. Regardless, it will be helpful to use in-situ comment and the "suggest" feature and point out the locations where UIS is missed. I too will look for them.


> 5. `test.cpp` [line 85](https://github.com/sigcpp/stl-lite/pull/18/commits/719b3fb18bfd9a37daee36a5f0dce211854408d0#diff-5e2c30b9e905f37296cccd6280d8fc73R85), do we need to throw an exception if the `name` is empty.

Possibly. I'll look into it.

test/array-test.cpp
```diff
@@ -30,41 +30,41 @@ void runTests()
 
 
 	//capacity
-	verify(!s.empty(), "s.empty()");
-	verify(s.size() == 3, "s.size()");
-	verify(s.max_size() == s.size(), "s.max_size()");
+	is_true(!s.empty(), "s.empty()");
```
cmcmone, line 33:
```suggestion
	is_false(s.empty(), "s.empty()");
```

test/array-test.cpp
```diff
@@ -30,41 +30,41 @@ void runTests()
...
+	is_true(s.size() == 3, "s.size()");
+	is_true(s.max_size() == s.size(), "s.max_size()");
 
-	verify(!p.empty(), "p.empty()");
-	verify(p.size() == 5, "p.size()");
-	verify(p.max_size() == p.size(), "p.max_size()");
+	is_true(!p.empty(), "p.empty()");
```
cmcmone, line 37:
```suggestion
	is_false(p.empty(), "p.empty()");
```

test/driver.cpp
```diff
@@ -56,18 +59,71 @@ int main(int argc, char* argv[])
...
-	catch (const std::string& msg) {
-		logLine(msg.data());
-		if (options.fom == file_open_mode::no_file)
-			show_error(msg.data());
+	catch (const cmd_line_error& cle) {
+		show_error_and_usage(cle.what(), argv[0]);
+		return -5;
+	}
+	catch (const std::exception& e) {
+		show_error((std::string{ "Unexpected error: " } +e.what()).data());
```
cmcmone, line 69:
```suggestion
		show_error((std::string{ "Unexpected error: " } + e.what()).data());
```

test/driver.cpp
```diff
@@ -56,18 +59,71 @@ int main(int argc, char* argv[])
...
+	std::map<std::string, suite_runner_type> suites{
+		//add one entry for each test suite to run
+		//suite-name and suite-runner, both in braces
+		{"array-test", array_test}
+	};
+
+	//build a collection of suites to run if necessary
+	//options.suites_to_run is empty or a semi-colon delimited list of suite names
+	const auto suites_to_run = split(options.suites_to_run, ';');
+	auto run_all_suites = suites_to_run.empty();
```
cmcmone, line 101:
```suggestion
	auto run_all_suites{ suites_to_run.empty() };
```

test/driver.cpp
```diff
@@ -56,18 +59,71 @@ int main(int argc, char* argv[])
...
+
+	//check that the suites specified are actually defined
+	//this check is not required to run the suites, but is included to inform the user of the issue
+	//silently ignoring leaves the user unaware of the reason a specified suite doesn't run
+	if (!run_all_suites) {
+		auto end_suites = suites.cend();
```
cmcmone, line 107:
```suggestion
		auto end_suites{ suites.cend() };
```

test/driver.cpp
```diff
@@ -56,18 +59,71 @@ int main(int argc, char* argv[])
...
+		for (auto& suite_name : suites_to_run) {
+			if (suites.find(suite_name) == end_suites) {
+				assert(false);
+				throw invalid_option_value{ std::string{"suite "} +suite_name + " not defined" };
```
cmcmone, line 111:
```suggestion
				throw invalid_option_value{ std::string{"suite "} + suite_name + " not defined" };
```

test/driver.cpp
```diff
@@ -56,18 +59,71 @@ int main(int argc, char* argv[])
...
+			}
+		}
+	}
+
+	//run all suites or only the suites indicated in options
+	auto size = suites.size();
```
cmcmone, line 117:
```suggestion
	auto size{ suites.size() };
```

test/driver.cpp
```diff
@@ -56,18 +59,71 @@ int main(int argc, char* argv[])
...
+	auto begin = suites_to_run.cbegin(), end = suites_to_run.cend();
```
cmcmone, line 118:
```suggestion
	auto begin{ suites_to_run.cbegin()}, end{ suites_to_run.cend()};
```

test/options.h
```diff
@@ -24,12 +24,13 @@ enum class file_open_mode { no_file, new_file, overwrite, append };
 struct Options {
 	bool header{ true };
 	bool summary{ true };
-	std::string header_text{ "Running $cmd" };
+	std::string header_text{ "Running $suite" };
 	pass_report_mode prm{ pass_report_mode::indicate };
 	unsigned short fail_threshold = 0;
```
cmcmone, line 29:
```suggestion
	unsigned short fail_threshold{ 0 };
```

test/tester.cpp
```diff
@@ -15,123 +15,165 @@
 #include <sstream>
 #include <string>
 #include <climits>
+#include <cassert>
 
+#include "utils.h"
 #include "tester.h"
 
+
 static std::string headerText("Running tests");
```
cmcmone, line 24:
```suggestion
static std::string header_text("Running tests");
```

test/tester.cpp
```diff
@@ -15,123 +15,165 @@
...
 
-void setMaxFailThreshold()
+void set_max_fail_threshold()
 {
-	failThreshold = USHRT_MAX;
+	fail_threshold = USHRT_MAX;
 }
 
 
 static std::ostream* pOut{ &std::cout };
```
cmcmone, line 51:
```suggestion
static std::ostream* p_out{ &std::cout };
```

test/tester.cpp
```diff
@@ -15,123 +15,165 @@
...
-void setOutput(std::ostream& o)
+void set_output(std::ostream& o)
 {
 	pOut = &o;
```
cmcmone, line 54:
```suggestion
	p_out = &o;
```

test/tester.cpp
```diff
@@ -15,123 +15,165 @@
...
 }
 
 
-//number of tests run and failed
-static unsigned testsDone;
-unsigned getTestsDone()
+void log(const char* s)
 {
-	return testsDone;
+	*pOut << s;
```
cmcmone, line 60:
```suggestion
	*p_out << s;
```

test/tester.cpp
```diff
@@ -15,123 +15,165 @@
...
 }
 
 
-static unsigned testsFailed;
-unsigned getTestsFailed()
+void log_line(const char* s)
 {
-	return testsFailed;
+	*pOut << s << '\n';
```
cmcmone, line 66:
```suggestion
	*p_out << s << '\n';
```

test/tester.cpp
```diff
@@ -15,123 +15,165 @@
...
+static unsigned suites_run;
+void start_suite(const std::string& name)
+{
+	assert(!name.empty());
+	suite_name = name;
+	++suites_run;
+
+	//print a separator between test suites
+	if (tests_done_suite != 0)
+		*pOut << "\n\n";
```
cmcmone, line 91:
```suggestion
		*p_out << "\n\n";
```

test/tester.cpp
```diff
@@ -15,123 +15,165 @@
...
+
+	tests_done_suite = 0;
+	tests_failed_suite = 0;
+
+	//print header text after expanding macro $suite
+	if (!headerText.empty()) {
+		const std::string suite_macro{ "$suite" };
+		std::string suite_header{ headerText };
+		replace_all(suite_header, "$suite", name);
+		*pOut << suite_header << ":\n";
```
cmcmone, line 101:
```suggestion
		*p_out << suite_header << ":\n";
```

test/tester.cpp
```diff
@@ -15,123 +15,165 @@
...
+//helper for formatting output
+bool last_output_ended_in_linebreak{ false };
 
-	*pOut << message.str();
+
+//print a report for the suite
+void summarize_suite()
+{
+	if (!last_output_ended_in_linebreak)
+		*pOut << "\n\n";
```
cmcmone, line 114:
```suggestion
		*p_out << "\n\n";
```

test/tester.cpp
```diff
@@ -15,123 +15,165 @@
...
+
+	*pOut << "Tests completed: " << tests_done_suite << '\n';
+	*pOut << "Tests passed: " << tests_done_suite - tests_failed_suite << '\n';
+	*pOut << "Tests failed: " << tests_failed_suite << '\n';
```
cmcmone, line 118:
```suggestion
	*p_out << "Tests completed: " << tests_done_suite << '\n';
	*p_out << "Tests passed: " << tests_done_suite - tests_failed_suite << '\n';
	*p_out << "Tests failed: " << tests_failed_suite << '\n';
```

test/tester.cpp
```diff
@@ -15,123 +15,165 @@
...
-	*pOut << "Tests failed: " << testsFailed << '\n';
+	*pOut << "Suites run: " << suites_run << '\n';
+	*pOut << "Tests completed: " << tests_done_total << '\n';
+	*pOut << "Tests passed: " << tests_done_total - tests_failed_total << '\n';
+	*pOut << "Tests failed: " << tests_failed_total << '\n';
 
-	if (testsFailed > failThreshold)
-		*pOut << "Tests stopped after " << testsFailed << " failure(s)\n";
+	if (tests_failed_total > fail_threshold)
+		*pOut << "Tests stopped after " << tests_failed_total << " failure(s)\n";
```
cmcmone, line 141:
```suggestion
		*p_out << "Tests stopped after " << tests_failed_total << " failure(s)\n";
```

test/tester.cpp
```diff
@@ -15,123 +15,165 @@
...
+		if (!last_output_ended_in_linebreak)
+			message << '\n';
+		message << "Test# " << tests_done_suite << ": FAIL (" << hint << ")\n";
+		last_output_ended_in_linebreak = true;
+
+		if (tests_failed_total > fail_threshold)
+			throw message.str();
+	}
+
+	*pOut << message.str();
```
cmcmone, line 178:
```suggestion
	*p_out << message.str();
```

test/utils.cpp
```diff
@@ -0,0 +1,26 @@
...
+#include <string>
+#include <string_view>
+
+#include "utils.h"
+
+//replace all instances of a substring with a new substring
+void replace_all(std::string& str, const std::string& substr, const std::string& new_substr)
+{
+	auto pos = str.find(substr);
+	auto substr_size = substr.size(), new_substr_size = new_substr.size();
```
cmcmone, line 23:
```suggestion
	auto substr_size{ substr.size() }, new_substr_size{ new_substr.size() };
```

test/driver.cpp
smurthys: Thanks for the suggestion, but would please state the issue? I'm unable to see it.

test/driver.cpp
smurthys: I'm glad you noticed this issue too. However, at least in my environment, formatting the document oddly remove that space between `+` and `suite_name`. Any idea what might be going on?

test/tester.cpp
smurthys: The convention is to use `p` prefix for pointers. Here, `p` does not represent a word. However, I recommend we do a survey of other code bases to see what people do in this case. Until then, I recommend not making this change.

smurthys:
@cmcmone Thanks for the suggestions. I've accepted many and left notes with others. Also, when there are many issues related to the same variable, it is better to just write a single overall comment about it instead of creating several in situ comments. I make this suggestion to make it easier for everyone.

smurthys:
I pushed a few changes that simplify the addition of new test suites as well as the process of using verifiers.  Much of the changes are in the new files `verifiers.h`, `suites.h`, and `suites.cpp`.

**To add a new test suite to the collection of suites to run:** Go to the  last section in the file `suites.cpp` and add an entry using the TEST_SUITE macro.

**To use verifiers in a test suite:** Include `verifiers.h` in any file where you need to call a verifier. Please do **NOT** include `tester.h` anymore.

smurthys:
> 5\. `test.cpp` [line 85](https://github.com/sigcpp/stl-lite/pull/18/commits/719b3fb18bfd9a37daee36a5f0dce211854408d0#diff-5e2c30b9e905f37296cccd6280d8fc73R85), do we need to throw an exception if the `name` is empty.

@cmcmone I concluded that even the assertion is unnecessary because the tester can run just fine no matter what the suite name is. Thus, I removed the assertion (and thus no need for error checking as well).

smurthys:
@sigcpp/stl-lite I am done making changes for now. It will be nice if one of you used the changes. Just in case you need it, here is a simple way to test if the changes in this PR work:

1. Assuming you have a local branch with your own test_suite, commit changes to the local repo (no need to push to origin)
2. Pull and locally check out the branch `multi-suite-simple` (the object of this PR)
3. Merge your branch from Step 1 to your local `multi-suite-simple`  
4. Change include `tester.h` in your own test suite code to include `verifiers.h`
5. Add your test suite to the suites collection in `suites.cpp`

It will be helpful if at least one of you is able to do this before today's call. I'm happy to get on a call with you between now and 2pm if it helps. I can also be available after 3pm. Thank you.

test/array-test.cpp
```diff
@@ -121,4 +121,4 @@ void array_test()
 	for (std::size_t idx = 0; idx < m.size() && swapTest; ++idx)
```
Ajetski, line 121:
```suggestion
	for (std::size_t idx{ 0 }; idx < m.size() && swapTest; ++idx)
```

test/driver.cpp
```diff
@@ -81,32 +82,23 @@ int main(int argc, char* argv[])
...
-	std::unordered_map<std::string, suite_runner_type> suites{
-		//add one entry for each test suite to run
-		//suite-name and suite-runner, both in braces
-		//not necessary, but recommend using the suite_runner function's name as the suite name
-		{"array_test", array_test}
-	};
-
-	//build a collection of suites to run if necessary
+	//retrieve all test suites defined
+	auto suites = get_test_suites();
```
Ajetski, line 89: Would it make more sense to use uniform initialization syntax here (and in the lines below)?

test/driver.cpp
Ajetski: It seems that @cmcmone is suggesting a formatting change; adding a space after the concatenation operator for clarity and readability. @smurthys

test/driver.cpp
smurthys: I believe I mentioned this in response to an earlier comment, but I expand on it here:

There is no benefit to using UIS with auto typed data.  By definition, auto chooses the exact type of the initializer and thus using UIS is pointless.

In general, people don't like UIS because it causes visual clutter and obscures the initial value, and makes it especially hard to comprehend what is going on when the initializer is a non-trivial expression. Thus, we should use UIS where we can't (and shouldn't) use auto.

test/array-test.cpp
smurthys: In general, traditionally people don't use UIS in for loop initialization, mostly because it makes it really hard to read code. However, IMO it is better to use them there as well. 

That said, the proposed change to `array-cpp` is not aligned with the goals of this PR. Yet, I would normally accept the suggestion, but I see that file needs many other (related) changes as well. So, I propose to delay making the change so an not to unnecessarily pollute the change list for this PR.

test/driver.cpp
smurthys: Please see my response [here](https://github.com/sigcpp/stl-lite/pull/18#discussion_r439748522).

smurthys:
I just pushed changes with variable-renaming and comment edits I made during the call earlier this afternoon. I also added checks for duplicate test-suite name and other errors when adding a test suite.

smurthys:
I just pushed the following two changes:

- Reject empty test suite name (though the macro does not permit an empty suite name)
- Define and use enum for error codes in `driver.cpp`

I expect these to be the last changes I expect to initiate on this PR. Of course, I'll address issues you report. I would appreciate being able to close this PR tonight. Thank you for your assistance.

smurthys:
Thanks @cmcmone for your review. It will be nice to format test summary as a table, and you are right that that task is not related to this PR or milestone. I recommend creating an issue so we can prioritize and track.

Any cmd-line parameter containing a has a space needs to be quoted if that item has to be treated as one value. This is an OS requirement. So, the OS considers the cmd-line `-run array_test; string_test` as having three parts `-run`, `array_test`, and `string_test`. Thus, the error message of "incorrect number of options" is correct. However, the cmd-line `-run "array_test; string_test"` would result in two parts `-run` and `array_test; string_test`.

That said, as far as the `-run` option goes, you need to be careful about the space in the option value. The value `array_test; string_test` asks to run a suite whose name begins with a space, and a suite with that exact name must exist. In theory, it is possible to have such suite names, but not with the `TEST_SUITE` macro.

A solution would be to trim leading and trailing spaces before looking up suite names in the collection, but I worry about spending valuable resources (time) on addressing extreme/unlikely cases. However, I also think it is unwise to not document or handle this issue at an appropriate time.

I propose you add an issue describing this problem and link to this (my) comment.

smurthys:
I updated the [cmd-line documentation](https://github.com/sigcpp/stl-lite/wiki/Command-line-interface-for-the-test-driver) to account for the notion of suites. It turned out to be lot of changes. 😮 

BTW, I'm not happy with the background of the image at the start of the doc. I don't know why the background has a mesh, but I am not able to attend to that at this moment.

cmcmone:
> ...I don't know why the background has a grid....

@smurthys I don’t quite understand, I don’t see any grid. :confused:

smurthys:
Compare the gray background in the image (the one with annotations) with the background in the Examples section. The background in the examples are smooth, whereas the background in the image is **not** as smooth. 

I believe the image's background was much smoother in the previous version, and @Ndemco had created that image. Clearly, I'm doing something wrong in how I created the image.

All this said, the image background is not a big issue. I am willing to live with it, at least for now.

cmcmone:
@smurthys like this?
image
[cmd-line.doc.example.pptx](https://github.com/sigcpp/stl-lite/files/4776928/cmd-line.doc.example.pptx)

smurthys:
Yes @cmcmone. The [image you show](https://github.com/sigcpp/stl-lite/pull/18#issuecomment-643797762) does have a smooth background even with the annotations.

cmcmone:
Set by `shape outline` -> `theme colors`.

image

smurthys:
Thanks @cmcmone. I'll try it next time. In the meantime, I used the image you shared in your earlier comment.

PR #23, opened 2020-06-15
The initial version of string_view test

cmcmone:
This PR is a test of the string_view of the C++ standard library.

The test content includes:
1. Construction and assignment
2. Capacity
3. Non-member comparison functions
4. Element data and access
5. Iterator support
6. Modifiers
7. String operations
8. Searching

**References:**
- [N4659: Final working draft of C++17 ](https://timsong-cpp.github.io/cppwp/n4659/string.view)
- [cppreference.com](https://en.cppreference.com/w/cpp/string/basic_string_view)


smurthys:
Thanks @cmcmone for creating the PR. I'm reviewing it now.

Quickly, the file `test-sql-lite.sln` is extra. Please remove it from this branch.

BTW, thanks for addressing issue #25.

test/string_view-test/string_view-test.cpp
```diff
@@ -0,0 +1,343 @@
+/*
+* string_view-test.cpp
+* Sean Murthy, Ray Chen
```
smurthys, line 3: I suggest placing your name first because you did quite a bit of work on this.

```suggestion
* Ray Chen, Sean Murthy 
```

test/string_view-test/string_view-test.cpp
```diff
@@ -0,0 +1,343 @@
...
+	//1. Construction and assignment
+	//2. Capacity
+
+	// Default constructor. Constructs an empty string_view. 
+	string_view empty_sv;
+	is_zero(empty_sv.size(), "empty_sv.size()");
+	is_true(empty_sv.length() == empty_sv.size(), "empty_sv.length() = empty_sv.size()");
+	is_true(empty_sv.empty(), "empty_sv.empty()");
+
+	//constructs a string_view with const char
```
smurthys, line 33: `"hello world"` is a `const char` array

```suggestion
	//constructs a string_view with const char[]
```

test/string_view-test/string_view-test.cpp
```diff
@@ -0,0 +1,343 @@
...
+	is_true(f_sv.length() == f_sv.size(), "f_sv.length() = f_sv.size()");
+	is_false(f_sv.empty(), "!f_sv.empty()");
+
+	//Copy constructor. Constructs a view of the same content as other.
+	string_view r_sv(f_sv);
+	is_true(r_sv.size() == f_sv.size(), "r_sv.size() = f_sv.size()");
+	is_true(r_sv.length() == r_sv.size(), "r_sv.length() = r_sv.size()");
+	is_true(r_sv.empty() == f_sv.empty(), "r_sv.empty() = f_sv.empty()");
+
+	//Assigns a view(replaces the view with that of view).
```
smurthys, line 51: Keep comment simple

```suggestion
	//assignment
```

test/string_view-test/string_view-test.cpp
```diff
@@ -0,0 +1,343 @@
...
+	is_true(sv.length() == sv.size(), "sv.length() = sv.size()");
+	is_false(sv.empty(), "!sv.empty()");
+
+	//Constructs a view of the first 5 characters of the character array
+	string_view f_sv("hello, world!", 5);
+	is_true(f_sv.size() == 5, "f_sv.size()");
+	is_true(f_sv.length() == f_sv.size(), "f_sv.length() = f_sv.size()");
+	is_false(f_sv.empty(), "!f_sv.empty()");
+
+	//Copy constructor. Constructs a view of the same content as other.
```
smurthys, line 45: Keep comments simple

```suggestion
	//copy constructor
```

smurthys:
@cmcmone Thanks again for the contribution. I know these tests were developed over a period of time since late April or so, and I recall our deliberations. I appreciate the effort and thought you have put into the tests.

Please consider the following questions/observations/suggestions. All items are about the file [`string_view-test.cpp`](https://github.com/sigcpp/stl-lite/blob/97c8aef1d9ec4680e72ae722e687a2c5e3014ff7/test/string_view-test/string_view-test.cpp).

- L11: The prologue comment should state the file's purpose. The next two lines are fine
- L46: What is the significance of the variable name `r_sv`. It does not have anything to do with the test case here (copy ctor). The prefix r makes me expect something about "reverse".
- L34, L40, and elsewhere: strive to use UIS, especially in initializing various string_view objects
- L40, L46, L52: test cases beginning on these lines should also test if `data()` function returns the same value as that of the source.
- It is better to create a `const char` array for `hello world!` at about L34 and use the named array throughout because there are too many occurrences of the same literal. The many occurrences can easily cause bugs if one of them somehow misspelled. (Such a named array is defined on L102)
- L59-L63: Change variable names s, t, p, cs1, and cs2 to something that shows the data they hold so that later code is easier to follow. After a while, we forget what s, t, etc. stand for.
- L65-L67: Use a string such `s == t` even in the hint instead of `s = t`. This observation applies in many tests before this line as well
- L105: Not sure what the comment "normal" means. Also, add a blank line before that comment. This happens later as well (L124). What does the comment "fixed length" mean? Are these comments misplaced and they should each be three lines above where they are now?
- I am unable to see any difference between the test case "normal" and "fixed length". Clearly I'm missing something
- normal, fixed-length, reference: It is unclear what test cases they represent. At the least, these sections are missing comment. it is also possible at least some of these tests are redundant, or can be merged with constructor tests earlier.
- L156: I am fairly convinced the comment //assignment is misplaced

I will stop here and review from iterators after I hear your response. My main concern right now is that L28-170 contain some redundant tests, and the test cases aren't clear. Also, I recommend splitting the tests into two or more functions because it becomes hard to follow after a while if the same function has too many lines.

I'm happy to answer questions in today's call.

cmcmone:
Thanks, @smurthys, I resolved some of the suggestions and keep moving on.

Ndemco:
@cmcmone On L41 - L45 where you construct a `string_view` of the first five characters of a character array: perhaps you should test that the value of `f_sv` is `"hello". You test that `f_sv` size is 5, but we do not know which 5 characters the copy constructor copied. There could be a bug where, for some reason, the copy constructor copied characters [2 - 6] and we wouldn't know.

cmcmone:
Thanks, @smurthys  for the review:

> L11: The prologue comment should state the file's purpose. The next two lines are fine

Rewritten prologue comment.

> L46: What is the significance of the variable name r_sv. It does not have anything to do with the test case here (copy ctor). The prefix r makes me expect something about "reverse".

Change the variable name `r_sv` to `c_sv`.

> L34, L40, and elsewhere: strive to use UIS, especially in initializing various string_view objects

These variables are initialized using the constructor, so I did not use UIS

> L40, L46, L52: test cases beginning on these lines should also test if `data()` function returns the same value as that of the source.

L28-L57 is to test `construction`, `assignment`, and `capacity`, so I did not test the `data()`. I tested `data()` in L102-L176

> It is better to create a const char array for `hello world!` at about L34 and use the named array throughout because there are too many occurrences of the same literal. The many occurrences can easily cause bugs if one of them somehow misspelled. (Such a named array is defined on L102)

fixed it up.

> L59-L63: Change variable names s, t, p, cs1, and cs2 to something that shows the data they hold so that later code is easier to follow. After a while, we forget what s, t, etc. stand for.

Change variable names s, t, p, cs1, and cs2 to ua(uppercase a), la(lowercase a), aua(another uppercase a), csA(C-string A), and csB(/C-string B)

> L65-L67: Use a string such s == t even in the hint instead of s = t. This observation applies in many tests before this line as well

fixed it up.

> L105: Not sure what the comment "normal" means. Also, add a blank line before that comment. This happens later as well (L124). What does the comment "fixed length" mean? Are these comments misplaced and they should each be three lines above where they are now?

fixed it up. BTW, `normal` means `default constructor and data test`, `fixed length` means `constructs a view of the first count characters of the character array and data test`.

> I am unable to see any difference between the test case "normal" and "fixed length". Clearly I'm missing something

The only difference is that they are initialized differently, one using the default constructor and one using "fixed length".

> `normal,` `fixed-length`, reference: It is unclear what test cases they represent. At the least, these sections are missing comment. it is also possible at least some of these tests are redundant, or can be merged with constructor tests earlier.

Possibly. We can discuss this in meeting.

> L156: I am fairly convinced the comment //assignment is misplaced

fixed it up.

test/string_view-test/string_view-test.cpp
```diff
@@ -0,0 +1,349 @@
...
+*
+* Attribution and copyright notice must be retained.
+* - Attribution may be augmented to include additional authors
+* - Copyright notice cannot be altered
+* Attribution and copyright info may be relocated but they must be conspicuous.
+*
+* Test the string_view of the C++ standard library and 
+* provide a test benchmark for subsequent implementation of string_view.
+* - see C++17 [string.view]
+*	https://timsong-cpp.github.io/cppwp/n4659/string.view
```
smurthys, line 14: Simplify prologue comment

```suggestion
* Unit tests for string_view
* - see C++17 [string.view]
* - https://timsong-cpp.github.io/cppwp/n4659/string.view
```

test/string_view-test/string_view-test.cpp
```diff
@@ -0,0 +1,349 @@
...
+#include "../verifiers.h"
+
+void string_view_test()
+{
+	using std::string_view;
+
+	//1. Construction and assignment
+	//2. Capacity
+
+	// Default constructor. Constructs an empty string_view.
```
smurthys, line 28: I strongly recommend including test for `data` member in each of the following test cases: default ctor, copy ctor, the two custom ctors, and assignment.

test/string_view-test/string_view-test.cpp
```diff
@@ -0,0 +1,349 @@
...
+
+	//constructs a string_view with const char[]
+	const char c_hw[]{ "hello, world!" };
+	string_view sv(c_hw);
+	is_true(sv.size() == 13, "sv.size()");
+	is_true(sv.length() == sv.size(), "sv.length() == sv.size()");
+	is_false(sv.empty(), "!sv.empty()");
+
+	//Constructs a view of the first 5 characters of the character array
+	string_view f_sv(c_hw, 5);
```
smurthys, line 42: I suggest using a non-C-string char array for this test case.

test/string_view-test/string_view-test.cpp
```diff
@@ -0,0 +1,349 @@
...
+	is_true(ua >= csA, "ua >= la");
+	is_true(ua <= csA, "ua >= la");
+	is_false(ua >= csB, "ua >= la");
+	is_true(ua <= csB, "ua >= la");
+
+
+	//4. Element data and access
+	is_true(empty_sv.data() == nullptr, "empty_sv.data()");
+
+	//const char* str_1 = "hello, world!";
```
smurthys, line 104: Revise comment

```suggestion
	//c_hw is defined earlier as const C-string "hello, world!"
```

test/string_view-test/string_view-test.cpp
```diff
@@ -0,0 +1,349 @@
...
+
+
+	//6. Modifiers
+	const char* sv_data{ sv.data() };
+	size_t sv_size{ sv.size() };
+	size_t n{ 1 };
+
+	sv.remove_prefix(n);
+	sv_size -= n;
+	is_true(sv.data() == sv_data + 1, "sv.remove_prefix().data()");
```
smurthys, line 239: Keep it simple. No need to use variable n here, especially because L239 hard codes sv_data + 1

test/string_view-test/string_view-test.cpp
```diff
@@ -0,0 +1,349 @@
...
+	is_true(sv.size() == sv_size, "sv.remove_prefix().size()");
```
smurthys, line 240: Simplify as follows:

```suggestion
	const char* data_expected sv.data()+1;
	size_t size_expected { sv.size() - 1 };

	sv.remove_prefix(n);
	is_true(sv.data() == data_expected, "sv.remove_prefix().data()");
	is_true(sv.size() == size_expected, "sv.remove_prefix().size()");
```

test/string_view-test/string_view-test.cpp
```diff
@@ -0,0 +1,349 @@
...
+	is_false(ua >= csB, "ua >= la");
+	is_true(ua <= csB, "ua >= la");
+
+
+	//4. Element data and access
+	is_true(empty_sv.data() == nullptr, "empty_sv.data()");
+
+	//const char* str_1 = "hello, world!";
+	//default constructor and data test
+	string_view sv_1(c_hw);
```
Ndemco, line 106: The comment says this tests the default constructor but I do not see the default constructor being tested.

test/string_view-test/string_view-test.cpp
```diff
@@ -0,0 +1,349 @@
...
+	sv_size -= n;
+	is_true(sv.data() == sv_data + 1, "sv.remove_prefix().data()");
+	is_true(sv.size() == sv_size, "sv.remove_prefix().size()");
+	
+	sv_data = sv.data();
+
+	sv.remove_suffix(n);
+	sv_size -= n;
+	is_true(sv.data() == sv_data, "sv.remove_suffix().data()");
+	is_true(sv.size() == sv_size, "sv.remove_suffix().size()");
```
smurthys, line 247: Simplify as follows. Also, the code looks a bit simpler and a bit more naturally sequenced if we test remove_suffix first,

```suggestion
	data_expected = sv.data();
	size_expected--;
	sv.remove_suffix(n);
	is_true(sv.data() == data_expected, "sv.remove_suffix().data()");
	is_true(sv.size() == size_expected, "sv.remove_suffix().size()");
```

smurthys:
@cmcmone Please consider the following questions/observations/suggestions in [string_view-test.cpp](https://github.com/sigcpp/stl-lite/blob/4b13ebf0f90ad3c153a82ded56008f13d177dec3/test/string_view-test/string_view-test.cpp):

- L61-L65: consider using variable names sv_A1, sv_A2, sv_a, zA, and zB. Also remove the end of line comments on these lines. They don't add any value
   - It might be better to use data with at least 2 chars instead of a single character; perhaps one sv with a single char will do
   - Need to also include an empty sv in the test 
- L74 and L82-L85: the hint text is incorrect. I believe it is incorrect in a couple of other places too
- L106-L173: replace with a loop to test subscript op against a C-string (like in the iterator test); and a loop to test at function against the same C-string; test front and back as you presently are. All this test with just one sv created from a C-string. No need to test [] and at for many sv instances.  Also you will have moved the test for data to the start of the suite.
-  L188, L197,  etc. move the declaration `size_t i` to for loop to avoid errors
- L194 and L 216: add a comment to say why these are being done again: to test const forward iterator and const reverse iterator
- L204-L205: change the order of declarations: it is better to first declare the sv and then the char array that says what is expected
- L 267: what is `f_sv`? It is impossible to remember what that variable is/has this far away from its original declaration.
- L267-L277: Missing test to see if `copy` worked as expected before testing exceptional cases. The same is true for `substr` later in the code
- L268: I'm unable to follow why/what is being done here. The entire code in that region begs a comment.
- L267: Sudden shift from nominal tests to exceptional tests. Please seriously consider my earlier advice of splitting the code into functions. The function is toooooo long now.
- FYI, I have **not** reviewed the tests for copy and substr because they are hard to follow.
- I really like the tests for comparison and searching

smurthys:
> > L34, L40, and elsewhere: strive to use UIS, especially in initializing various string_view objects
> 
> These variables are initialized using the constructor, so I did not use UIS

Even ctors take UIS.

> > L105: Not sure what the comment "normal" means. Also, add a blank line before that comment. This happens later as well (L124). What does the comment "fixed length" mean? Are these comments misplaced and they should each be three lines above where they are now?
> 
> fixed it up. BTW, `normal` means `default constructor and data test`, `fixed length` means `constructs a view of the first count characters of the character array and data test`.

Code and comments should be clear without need for outside explanation. Of course, it could refer to documentation, but that should be only to communicate details.

cmcmone:
Thanks, @Ndemco for review.

> On L41 - L45 where you construct a string_view of the first five characters of a character array: perhaps you should test that the value of f_sv is "hello".

Yes, you are right! I didn't test the value here because we need to test `==`  is correct first. so, I tested the `data` in [L123](https://github.com/sigcpp/stl-lite/pull/23/commits/71a0b1b3aa3e467811768285e7f58ee949fd68b6#diff-c7c4d56d7521f7fb4bedfac8b8740b79R123), after the `non-member comparison functions`.

> There could be a bug where, for some reason, the copy constructor copied characters [2 - 6] and we wouldn't know.

I don't know either, maybe more detail will help.

test/string_view-test/string_view-test.cpp
cmcmone: I'm still a little confused. Shouldn't we test the `data` after the `non-member comparison functions` like `==` is correct? Because we need `==` to judge whether the value of data is as expected, such as `sv.data() == somevalue` or `sv_2[0] == somechar`.

Ndemco:
@cmcmone 

> I don't know either, maybe more detail will help.

I'm not saying there is a bug, but let's say there was a bug in our stl-lite `string_view` implementation where something like `string_view f_sv(c_hw, 5)` wasn't properly setting `f_sv` to the first five characters of the underlying character array `[c_hw[0] - c_hw[4]`, and instead setting `f_sv` to `[c_hw[1] - c_hw[5`]. In this case, `f_sv` would be `ello `  with a space at the end. You want to test that `f_sv == "hello"`.

test/string_view-test/string_view-test.cpp
smurthys: Perhaps I am mistaken, but consider this code. This code is testing if `data()` does the right thing. I don't understand the dependency you say exists with some test later.

PS: I'm away and expect to return by 9pm

```cpp
const char z[]{"he"};
std::string_view sv{z};
bool b = sv.data() == z;
```

test/string_view-test/string_view-test.cpp
cmcmone: add a non-C-string char array for this test case. 2830432

test/string_view-test/string_view-test.cpp
cmcmone: yes, in your code. the `sv.data() == z` return a bool value. but how do we know the return value is correct? so, in this case, it dependency the operator `==`, the return value is correct only when operator `==` is overloaded correctly. So I have to test `non-member comparison functions` first.

test/string_view-test/string_view-test.cpp
cmcmone: Thanks @Ndemco It's my mistake.

test/string_view-test/string_view-test.cpp
```diff
@@ -0,0 +1,349 @@
...
+	//swap(exchanges the values of *this and s.)
+	string_view o_sv("lunch!");
+	string_view s_sv("dinner");
+	char owExpected[]{ "lunch!" };
+	char swExpected[]{ "dinner" };
+	o_sv.swap(s_sv);
+	bool swapTest = true;
+	for (std::size_t idx = 0; idx < o_sv.size() && swapTest; ++idx)
+		swapTest = o_sv[idx] == swExpected[idx] && s_sv[idx] == owExpected[idx];
+	is_true(swapTest, "o_sv.swap(s_sv)");
```
Ndemco, line 261: This is very hard to follow as the variable names are confusing. Is there a reason you can't just use `string_view`'s `==` operator here?

```suggestion
	//swap(exchanges the values of *this and s.)
	string_view sv_1("lunch!");
	string_view sv_2("dinner");
	char lunch[]{ "lunch!" };
	char dinner[]{ "dinner" };
	sv_1.swap(sv_2);
	is_true(sv_1 == dinner && sv_2 == lunch, "sv_1.swap(sv_2)")
```
 I understand this relies on the implemented `==` function to be correct which perhaps we shouldn't assume since it is one of the things being tested. If not using the `==` operator, perhaps write a function that compares a `string_view` to a `c-string` and returns true if the values are equal to make it more readable.

test/string_view-test/string_view-test.cpp
smurthys: I feel there is a misunderstanding. The `==` operator here is **not** sv's overloaded. Instead, it is the built in `==` operator and is just comparing two pointers. The thing about whether the content at the location `data()` returns is tested later with iterators, subscript, and `at`. That is the only way we can test if the data content is correct. Except, here we only want to know if the pointer we gave sv at creation is the pointer that `data()` returns.

smurthys:
> I'm not saying there is a bug, but let's say there was a bug in our stl-lite `string_view` implementation where something like `string_view f_sv(c_hw, 5)` wasn't properly setting `f_sv` to the first five characters of the underlying character array `[c_hw[0] - c_hw[4]`, and instead setting `f_sv` to `[c_hw[1] - c_hw[5`]. In this case, `f_sv` would be `ello ` with a space at the end. You want to test that `f_sv == "hello"`.

The point is valid @Ndemco. However, content test is done later on when iterators, subscript, and `at` are tested. Those are the only ways to test content.

test/string_view-test/string_view-test.cpp
smurthys: I agree with @Ndemco. A robust test here needs to be: `sv1.data() == dinner && sv1.size() == sizeof(dinner) && sv1 == dinner;`. Of course, three more conjunctive predicates are needed against sv2 for the complete test.

cmcmone:
>>> L34, L40, and elsewhere: strive to use UIS, especially in initializing various string_view objects

>>These variables are initialized using the constructor, so I did not use UIS

>Even ctors take UIS.

Even other ctors?
such as:

```c++
//constructs a view of the first 5 characters
string_view f_sv{ c_hw, 5 };
//copy constructor
string_view c_sv{ f_sv };
//assignment
string_view a_sv{ c_sv };
```
I think using `()` is easier to understand this is ctors case.

cmcmone:
Thanks @Ndemco 
> > I don't know either, maybe more detail will help.
> 
> I'm not saying there is a bug, but let's say there was a bug in our stl-lite `string_view` implementation where something like `string_view f_sv(c_hw, 5)` wasn't properly setting `f_sv` to the first five characters of the underlying character array `[c_hw[0] - c_hw[4]`, and instead setting `f_sv` to `[c_hw[1] - c_hw[5`]. In this case, `f_sv` would be `ello ` with a space at the end. You want to test that `f_sv == "hello"`.

There should be no bugs here because we are testing the C++ standard library. I review my code and it got the correct value. so, I don't know what happened in this case.

image

[string.view.cons](https://en.cppreference.com/w/cpp/string/basic_string_view/basic_string_view)

smurthys:
> Even other ctors?
> such as:
> 
> ```c++
> //constructs a view of the first 5 characters
> string_view f_sv{ c_hw, 5 };
> //copy constructor
> string_view c_sv{ f_sv };
> //assignment
> string_view a_sv{ c_sv };
> ```
> 
> because this is a test for ctors, so I think using `()` is easier to understand.

Yes, why not use UIS with all ctors? The purpose of a ctor's is unrelated to using UIS. Also UIS is just as understandable as ().

test/string_view-test/string_view-test.cpp
cmcmone: > This is very hard to follow as the variable names are confusing. Is there a reason you can't just use string_view's == operator here?

Thanks @Ndemco for the question, sorry about that my code is hard to read.

The reason I use for loop comparison here is because we can‘t just compare whether the pointers are the same. If the want sure that a `string_view` is equal to a `C-string`, then each character in this `string_view` and `C-string` must be equal.

But I think your writing is also correct, because when you use the operator `==` to compare, it also compares each character because of `string_view` overloads this operator. 

@smurthys may have a better explanation.

cmcmone:
> @cmcmone I worry I'm not looking right. I see the commit message says new filters were created in the VS solution, but I don't really see any actual filters under "Source Files" in Solution Explorer. I do however see that the test suites are correctly moved to their own file-system folders which was done in another commit. No problem with that.
> 
> What am I missing?

Yes,  I didn't create a new filter. I just change `#include "verifiers.h"` to `#include "../verifiers.h"` in `string_view-test.cpp` and VS automatic add this
```
    <ClCompile Include="string_view-test\string_view-test.cpp">
      <Filter>Source Files</Filter>
    </ClCompile>
```
to the `test-stl-lite.vcxproj.filters` file, and it's work. so I think this is your means `new filter`. I may have misunderstood.

cmcmone:
> > Even other ctors?
> > such as:
> > ```c++
> > //constructs a view of the first 5 characters
> > string_view f_sv{ c_hw, 5 };
> > //copy constructor
> > string_view c_sv{ f_sv };
> > //assignment
> > string_view a_sv{ c_sv };
> > ```
> > 
> > 
> > because this is a test for ctors, so I think using `()` is easier to understand.
> 
> Yes, why not use UIS with all ctors? The purpose of a ctor's is unrelated to using UIS. Also UIS is just as understandable as ().

If we use UIS for `assignment`, this will make it no different with copy constructor.

cmcmone:
> L61-L65: consider using variable names sv_A1, sv_A2, sv_a, zA, and zB. Also remove the end of line comments on these lines. They don't add any value
> * It might be better to use data with at least 2 chars instead of a single character; perhaps one sv with a single char will do
> * Need to also include an empty sv in the test

fixed, but I can't use `sv_a` because it was defined earlier. used `sv_a1` instead. ea066a791bedb67157f7dab81fcffc73e12dc629

> * L74 and L82-L85: the hint text is incorrect. I believe it is incorrect in a couple of other places too

fixed it up. 31c405b8c2687ad76ed90843faa8a01b4dfb8f90

> * L106-L173: replace with a loop to test subscript op against a C-string (like in the iterator test); and a loop to test at function against the same C-string; test front and back as you presently are. All this test with just one sv created from a C-string. No need to test [] and at for many sv instances.  Also you will have moved the test for data to the start of the suite.

***Not finished yet***

> * L188, L197,  etc. move the declaration `size_t i` to for loop to avoid errors

fixed it up. d7ce033dd7472b0914b815934c1bb74b73b72d01

> * L194 and L 216: add a comment to say why these are being done again: to test const forward iterator and const reverse iterator

I'm not sure the comment is meet. fac23f9258b171ce3c895d77958804e3acf879f5

> * L204-L205: change the order of declarations: it is better to first declare the sv and then the char array that says what is expected

fixed it up. 9319c5cdc1dda344f0cd120e1fe3eada9adf3672

> * L 267: what is `f_sv`? It is impossible to remember what that variable is/has this far away from its original declaration.

add a comment for `f_sv`  a4de6345d2f896c74a0ba9fab27967b82fd55459

> * L267-L277: Missing test to see if `copy` worked as expected before testing exceptional cases. The same is true for `substr` later in the code

add test before testing exceptional cases. 1d0f900c35b81d978116bbd0b7a1b7cd7ee3ae1a

> * L268: I'm unable to follow why/what is being done here. The entire code in that region begs a comment.

add a comment 51fbb1dfda5f99babd2830b16af67e095c9bfedf

> * L267: Sudden shift from nominal tests to exceptional tests. **Please seriously consider my earlier advice of splitting the code into functions**. The function is toooooo long now.

***Not finished yet***

Some problems didn't finish, I need to think about it and finish it later.

smurthys:
> If we use UIS for `assignment`, this will make it no different with copy constructor.

Of course. We wouldn't use UIS for assignment. However, the following line calls the copy ctor; not assignment: `string_view a_sv = c_sv;` Assignment is called only if an already constructed instance is assigned another instance.

smurthys:
@cmcmone FYI, I am looking over the recent changes. In the meantime, I like to share my appreciation for your perseverance. I hope the reviews are helpful in understanding key issues such as variable names, missing/misplaced comments, and redundant code. However, I think the code is getting better.

smurthys:
I just pushed a change to create test-suite specific filters under "Source Files" in the Visual Studio solution. This change fully fixes #25 

@cmcmone I fixed this issue because I notice you are using VS Code. It was easier for me to fix it than ask you to use VS 2019.

smurthys:
Building the current version (HEAD 67aa282) produces warnings in `string_view-test.cpp`. I recommend addressing the warnings before making any other change:

```console
1>string_view-test.cpp
1>C:\Git\sigcpp\stl-lite\test\string_view-test\string_view-test.cpp(306,15): warning C4389: '==': signed/unsigned mismatch
1>C:\Git\sigcpp\stl-lite\test\string_view-test\string_view-test.cpp(306,96): warning C4800: Implicit conversion from 'size_t' to bool. Possible information loss
1>C:\Git\sigcpp\stl-lite\test\string_view-test\string_view-test.cpp(306,96): message : consider using explicit cast or comparison to 0 to avoid this warning
1>C:\Git\sigcpp\stl-lite\test\predefined C++ types (compiler internal)(16,22): message : see declaration of 'size_t'
1>C:\Git\sigcpp\stl-lite\test\string_view-test\string_view-test.cpp(330,24): warning C4389: '==': signed/unsigned mismatch
1>C:\Git\sigcpp\stl-lite\test\string_view-test\string_view-test.cpp(330,94): warning C4800: Implicit conversion from 'size_t' to bool. Possible information loss
1>C:\Git\sigcpp\stl-lite\test\string_view-test\string_view-test.cpp(330,94): message : consider using explicit cast or comparison to 0 to avoid this warning
1>C:\Git\sigcpp\stl-lite\test\predefined C++ types (compiler internal)(16,22): message : see declaration of 'size_t'
```

smurthys:
I pushed changes to simplify iterator test (4b66847), swap test (b56d319), and member-access test (db07cce). Please review and opine. FYI, the number of tests has now come down to 146 (from 166).

In all the changes I made, I retained as much of the existing code and style, even though that leaves a few unsatisfactory things in the code. However, I believe addressing my earlier comments will address the things that are left in.

PS: I'm forced to suspend working on this review at this time in order to attend to other work. I'll monitor messages, but I'm not sure if/when I can reply.

cmcmone:
> I just pushed a change to create test-suite specific filters under "Source Files" in the Visual Studio solution. This change fully fixes #25

So `create new filters` means to add the path to `<Filter>Source Files\string_view-test</Filter>`? What is the effect of doing this?

cmcmone:
Thanks @smurthys for the push change

> I pushed changes to simplify iterator test ([4b66847](https://github.com/sigcpp/stl-lite/commit/4b66847a2cec7d6f738a4ab6aa8024f952149fed)), swap test ([b56d319](https://github.com/sigcpp/stl-lite/commit/b56d31911e40974da4cf92ef8b86e1f00885b6af)), and member-access test ([db07cce](https://github.com/sigcpp/stl-lite/commit/db07cce1c449fa9fcc0e9788dd5de59922016f7f)). Please review and opine. FYI, the number of tests has now come down to 146 (from 166).

For the swap test (b56d319)

```c++
bool swap_test = o_sv.data() == c_dinner && o_sv.size() == sizeof(c_dinner) - 1 && 
		other_sv.data() == c_lunch && other_sv.size() == sizeof(c_lunch) - 1;
```

can I just write like @Ndemco's code

```c++
bool swap_test = o_sv == c_dinner && other_sv == c_lunch;
```
because the operator `==` is overload and will compare both data and size.

cmcmone:
> Building the current version (HEAD [67aa282](https://github.com/sigcpp/stl-lite/commit/67aa282d36a35e87356e4b31364c5e80d505b3ea)) produces warnings in `string_view-test.cpp`. I recommend addressing the warnings before making any other change:
> 
> ```
> 1>string_view-test.cpp
> 1>C:\Git\sigcpp\stl-lite\test\string_view-test\string_view-test.cpp(306,15): warning C4389: '==': signed/unsigned mismatch
> 1>C:\Git\sigcpp\stl-lite\test\string_view-test\string_view-test.cpp(306,96): warning C4800: Implicit conversion from 'size_t' to bool. Possible information loss
> 1>C:\Git\sigcpp\stl-lite\test\string_view-test\string_view-test.cpp(306,96): message : consider using explicit cast or comparison to 0 to avoid this warning
> 1>C:\Git\sigcpp\stl-lite\test\predefined C++ types (compiler internal)(16,22): message : see declaration of 'size_t'
> 1>C:\Git\sigcpp\stl-lite\test\string_view-test\string_view-test.cpp(330,24): warning C4389: '==': signed/unsigned mismatch
> 1>C:\Git\sigcpp\stl-lite\test\string_view-test\string_view-test.cpp(330,94): warning C4800: Implicit conversion from 'size_t' to bool. Possible information loss
> 1>C:\Git\sigcpp\stl-lite\test\string_view-test\string_view-test.cpp(330,94): message : consider using explicit cast or comparison to 0 to avoid this warning
> 1>C:\Git\sigcpp\stl-lite\test\predefined C++ types (compiler internal)(16,22): message : see declaration of 'size_t'
> ```
fixed it up.

test/string_view-test/string_view-test.cpp
```diff
@@ -0,0 +1,390 @@
...
+
+	const char* a_ABC{ "ABC" };
+	is_positive(sv_a_to_e.compare(a_ABC), "sv_a_to_e.compare(str)");
+	is_zero(sv_A_to_E.compare(0, 3, a_ABC), "sv_empty.compare(0, 5, str)");
+	is_positive(sv_a_to_c.compare(a_ABC), "sv_empty.compare(0, 5, str, 5)");
+
+
+	//8. Searching
+	//    0         1 
+	//    01234567890123456
```
Ndemco, line 368: Perhaps I'm missing something but what are these comments for?

cmcmone: I fixed it. [see @smurthys's comment ](https://github.com/sigcpp/stl-lite/pull/23#commitcomment-38708357)

Ndemco: Ahh OK that makes sense. It is hard to know what those numbers are for without seeing that comment.

smurthys:
@cmcmone The tests are much more readable now. I feel we could have been further along than this if the code was broken down to functions early on, but better late than never.

I pushed some changes with minor changes: reorganization, comments, renaming, might have even fixed a bug (I think in `test_modifiers`). I made the change myself because they were various and it seemed more efficient to make the changes than describe them for you to make the changes.

Here are some observations on [string_view-test.cpp](https://github.com/sigcpp/stl-lite/blob/9b888346795e4ff5eb9b7a5d4e2d133b335786fa/test/string_view-test/string_view-test.cpp):
- Need an exception handling test for `at`
- I haven't understood the need for four different versions of `test_element_access_`. Why do we need four? What do the other three variation do that the first one does not do, other than use a different means to create the sv object? If the creation means is the only difference (that is all I see), I think we only need one of the four functions
- copy and substr tests are still somewhat hard to read
   - declare `len_expected` close to first point of use: remember the principle of locality of reference.
   - use `strncmp` instead of `strcmp`: strcmp requires C-string and is not safe
   - use `std::strncmp` (namespace qualify)
   - Add a space before the 3-line comment 
   - I like how the 3-line comment says what to expect, but it helps if the variable names used match what is in the comment
   - Add a comment before the try block to say what/why the test is: depending on how this comment is worded additional comments inside the try block and the first catch block may or may not be needed
   - The hint text in the verifiers used in the exception handlers are wordy
   - In general, use white space well

smurthys:
Thanks @cmcmone for the changes. I am reviewing them. The copy and substr tests are clearer, but at least the copy test could be improved. FYI, **I am making those changes right now**.

smurthys:
I just pushed changes to copy and substr tests. I also made a few minor comment edits and stuff. Please review to make sure I didn't mess up something.

smurthys:
I believe if we all can review my recent changes, and the code overall, we should be able to close this PR today.

cmcmone:
Thanks @smurthys for the change.
I reviewed and everything looks great! I think we are ready to close this PR. Thanks again @smurthys @Ndemco.

smurthys:
I renamed string_view variables in a couple of functions because there is no longer a need to use long and contrived names due to code being broken down into functions. Please eyeball the changes.

smurthys:
@Ndemco do you have any comments/thoughts on this PR?

smurthys:
Thanks again @cmcmone for your effort and perseverance on this PR.

Issue #25, opened 2020-06-16
Too many files in the project folder


smurthys:
Presently, there are about 17 files in the test folder and the number of files is only expected to grow. The issue is that files from the test driver and the test suites are in the same folder.

I propose placing the files for each test suite in its own sub-folder and also create a new filter under "Source Files"  in the Visual Studio solution. So, as of now, this proposal will cause three sub-folders: `driver-test`, `array-test`, `string_view-test`

The proposal also enables automation of determining and acting on test suites (with a certain organizational convention enforced)

smurthys:
@cmcmone has addressed this issue in commit 7745eec which is included in PR #23

PR #27, opened 2020-06-17
Test get_options for nominal behavior

Ndemco:
The commits in this PR begin the process of creating a unit test for the test driver. Following the [milestone 6](https://github.com/sigcpp/stl-lite/milestone/6) goals, this PR tests only the function **get_options** and only tests for **nominal behavior**.

The following changes have been made:
* Add `driver-test.cpp` to solution.
    * Create function `driver_test` that will call all of the modularized test functions.
    * Create function `test_get_options` which will be called by `driver_test`, and other necessary functions to test `get_options`.

* Add the test suite `driver_test` to the suites collection in `suites.cpp`

This PR also addresses issue #25 and creates a driver-test subfolder that driver-test-specfic files will be in, as well as creates a filter for these files in VS2019.


smurthys:
Thanks for the PR @Ndemco. I will look over as much as I can before this afternoon's call.

test/driver-test/driver-test.cpp
```diff
@@ -0,0 +1,312 @@
...
+void test_get_options();
+void reset_default_options(Options& o);
+void compare(const Options& received, const Options& expected, const std::string& test_case);
+std::vector<std::pair<std::string, Options>> generate_cmdline_option_pairs();
+
+void driver_test()
+{
+	test_get_options();
+}
+
```
smurthys, line 68: Have two empty lines between functions. Also, add a line of prologue comment to each function, unless the function's purpose is is really really obvious **to others**.

test/driver-test/driver-test.cpp
```diff
@@ -0,0 +1,312 @@
...
+void test_get_options()
+{
+	char* args[100];
+
+	const std::vector<std::pair<std::string, Options>> args_and_options = generate_cmdline_option_pairs();
+	std::size_t cmd_line_number{ 0 };
+	for (auto iter = args_and_options.begin(); iter != args_and_options.end(); ++iter)
```
smurthys: Curious about the choice to use iterator. Why not use range-for?

test/driver-test/driver-test.cpp
```diff
@@ -0,0 +1,312 @@
...
+	is_true(received.header_text == expected.header_text, (test_case + " header text").data());
+	is_true(received.prm == expected.prm, (test_case + " pass report mode").data());
+	is_true(received.fail_threshold == expected.fail_threshold, (test_case + " fail threshold").data());
+	is_true(received.fom == expected.fom, (test_case + " file open mode").data());
+	is_true(received.output_filepath == expected.output_filepath, (test_case + " output filepath").data());
+	is_true(received.command_name == expected.command_name, (test_case + " command name").data());
+	is_true(received.suites_to_run == expected.suites_to_run, (test_case + " suites to run").data());
+}
+
+std::vector<std::pair<std::string, Options>> generate_cmdline_option_pairs()
```
smurthys: At first look, this function interface seems more complicated than it should be. At the least, this function is easily improved by returning a map instead of a vector of pairs where each pair has a unique string "test case name" and the corresponding options. This is exactly what a map already does with less clutter.

test/driver-test/driver-test.cpp
```diff
@@ -0,0 +1,312 @@
...
+std::vector<std::pair<std::string, Options>> generate_cmdline_option_pairs();
+
+void driver_test()
+{
+	test_get_options();
+}
+
+void test_get_options()
+{
+	char* args[100];
```
smurthys, line 74: This declaration begs a comment. We know what it does. We don't know why this data exists, and why the bound of 100 is chosen.

smurthys:
Good start @Ndemco.

I have placed a few comments _in situ_, and have the following overall observation: The current setup is a bit complicated. I see `test_get_options` as containing the following conceptual steps for each test case:

1. Set cmd-line string
2. Set up expected options: copy a reference options struct that is already inited with the most common values and change the members appropriately in the copy. No need for function `reset_options`
3. Get a vector of strings after splitting the command line: simply call `split`
4. Copy data from every string in the vector to `args` array
5. Call `get_options`
6. Compare received and expected options

Steps 1 and 2 can be done in `test_get_options`; the other steps can be done in a function. So, `test_get_options` will essentially have just 3 steps for each test case. Depending on the number of lines to set up the expected options (at least 2 lines, may be 3-5, average 3.5 lines), each test case should take about 5.5 lines. 30 test cases should have ~165 lines. As comparison, `generate_cmdline_option_pairs` has ~210 lines, and may be get_options will get that long, but it will be significantly simpler.

This is my imagination. It is possible I have overlooked something, but I feel my approach is worth a try.

test/driver-test/driver-test.cpp
Ndemco: You are right, a range-for would be much better here.

test/driver-test/driver-test.cpp
Ndemco: @smurthys I thought about using a map, but I chose a vector because a vector gives me the ability to order by when the element was inserted. The dummy cmd-lines are grouped in a logical way that tests related cases together (eg. test single name-value options first, test edge cases last), I wanted the to run the tests in that same logical order. As far as I know, a std::map will sort by the value of the key, and a std::unordered_map does not guarantee any specific order of its elements, thus, iterating through a map to run the tests will not produce the order of tests I was looking for. A map could definitely de-clutter the code, but the look-up and insertion performance that is often the primary pro of using a map does not benefit us here and we would lose the logical order of tests run. Perhaps the pros of using a map outweigh the cons?

smurthys: Good perspective @Ndemco. A few observations: there is no benefit to running the tests in the order written unless there is dependency between tests (which isn't the case here). Even then, `std::map` guarantees order and you can even enforce a specific order.

This said, a vector of pairs or a map is unnecessary as I have outlined in my [other comment](https://github.com/sigcpp/stl-lite/pull/27#issuecomment-645542322). Using the approach in that suggestion is what really reduces clutter, complexity, and complications.

test/driver-test/driver-test.cpp
```diff
@@ -0,0 +1,325 @@
...
+		compare(o, iter->second, "cmd_line_" + std::to_string(cmd_line_number));
+		++cmd_line_number;
+	}
+}
+
+
+//resets given Options& parameter data members to the default values
+void reset_default_options(Options& o)
+{
+	o.header = true;
```
cmcmone:
```suggestion
	o.header{ true };
```
use UIS

cmcmone: [L63](https://github.com/sigcpp/stl-lite/pull/27/files#diff-a652f65fb5216681fe6cd07d6e6d05baR63)
[L66](https://github.com/sigcpp/stl-lite/pull/27/files#diff-a652f65fb5216681fe6cd07d6e6d05baR66)

test/driver-test/driver-test.cpp
```diff
@@ -0,0 +1,325 @@
...
+#include <iostream>
+#include <vector>
+#include <string>
+
+#include "../verifiers.h"
+#include "../options.h"
+#include "../tester.h"
+#include "../utils.h"
+
+void driver_test();
```
cmcmone: I think we don't need to declare `driver_test` function here

test/driver-test/driver-test.cpp
cmcmone: can just use `auto` instead of `std::vector<std::pair<std::string, Options>>` in L43? and L47 `std::vector<std::string>`

test/driver-test/driver-test.cpp
```diff
@@ -0,0 +1,325 @@
...
+//resets given Options& parameter data members to the default values
+void reset_default_options(Options& o)
+{
+	o.header = true;
+	o.summary = true;
+	o.header_text = "Running $suite";
+	o.prm = pass_report_mode::indicate;
+	o.fail_threshold = 0;
+	o.fom = file_open_mode::no_file;
+	o.output_filepath = "";
```
cmcmone: ~can just call the default constructor? no need to assign a null value to it. L68 and L70~ Edit: not right, same reason [@Ndemco's comment](https://github.com/sigcpp/stl-lite/pull/27#discussion_r441930853)

Ndemco:
@smurthys I've created a function that does steps 3 through 6 in your suggestion:

```
void parse_cmd_line_and_compare(char* args[], const std::string& cmd_line, 
const Options& expected, const std::string& test_case)
{
	std::vector<std::string> v = split(cmd_line, ' ');
	std::size_t size = v.size();
	for (std::size_t i = 0; i < size; ++i)
		args[i] = v[i].data();

	Options received = get_options(args, size);
	compare(received, expected, test_case);
}
```
Following steps 1 through 6 of your suggestion, I wrote this code into `test_get_options` and it seems to work as intended:

```
void test_get_options()
{
    char* args[100];

    Options o;
    o.command_name = "array_test";
	
    const std::string cmd_line_0("C:/Libraries/stl-lite/array_test.exe");
    Options expected = o;
    parse_cmd_line_and_compare(args, cmd_line_0, expected, "cmd_line_0");
}
```
The function is doing a lot of different things which makes it hard to give it an appropriate name, which then makes it hard to know exactly what the function is doing without looking at the code, but I do agree this is probably a better way to go about it. `test_get_options` is just one part of the `driver_test`, it does not make sense and is distracting to have a 200+ line function in `driver.cpp` just to be called in `test_get_options`.

test/driver-test/driver-test.cpp
Ndemco: I don't think UIS is appropriate or will work here since `o.header` was initialized when `Options o` was created.

smurthys:
@Ndemco I will comment as best as I understand from the short description, but I need to see things in context which is possible if you push the changes. At a high level, I don't think the proposed function `parse_cmd_line_and_compare` needs the `args` parameter. That is an unnecessary dependency.

> but I do agree this is probably a better way to go about it
I see you are still not fully convinced that this approach is simpler and better than using a map or a large vector of pairs. 😄 

> `test_get_options` is just one part of the `driver_test`, it does not make sense and is distracting to have a 200+ line function in `driver.cpp` just to be called in `test_get_options`.

I'm afraid I don't follow. You do have the choice to rename functions and the choice to partition a function into other functions. There is also the option of spreading code across files. Indeed, creating and naming naming abstractions is key part of software design. Plus, none of this code will be in `driver.cpp`, and testing `get_options` for nominal behavior is just one scenario. There are so many other scenarios that need testing.

Ndemco:
OK. I have just pushed the changes.

> I'm afraid I don't follow.

Sorry, I don't think last part of my comment was very clear. What I meant was, `driver-test.cpp` (I mistakenly said `driver.cpp`) is going to have a lot more to it by the time `driver_test` is complete and it does not make sense to have a 200+ line function (`generate_cmd_line_option_pairs`) simply to be called in `test_get_options`. I agree with your suggestion that moving the generating of cmd_line strings to `test_get_options` and simplifying the behavior is easier to understand and more readable.

smurthys:
Thanks for the clarification @Ndemco and for the changes.

> it does not make sense to have a 200+ line function (`generate_cmd_line_option_pairs`) simply to be called in `test_get_options`

Wasn't there a 200+ line function before this refactoring as well, and get_tesp_options was even then just calling a 200+ line function. We may still have a 200+ line function but the function interface as well as the entire structure is significantly simpler. If the concern is that a single function has 200 lines, and that is a valid concern, nothing prevents us from splitting the function at logical boundaries. In fact, we just went through that process in PR #23.

test/driver-test/driver-test.cpp
```diff
@@ -0,0 +1,306 @@
+/*
+* driver-test.cpp
+* Nick DeMarco
```
smurthys: I believe including my name (second) is appropriate because I have made significant contribution to the design of and implementation of this file's contents.

test/driver-test/driver-test.cpp
```diff
@@ -0,0 +1,306 @@
...
+
+#include "../verifiers.h"
+#include "../options.h"
+#include "../tester.h"
+#include "../utils.h"
+
+void driver_test();
+void test_get_options();
+void compare(const Options& received, const Options& expected, const std::string& test_case);
+void parse_cmd_line_and_compare(char* args[], const std::string& cmd_line, const Options& expected, const std::string& test_case);
```
smurthys: The definition of function `parse_cmd_line_and_compare` does not have an `args` parameter.

test/driver-test/driver-test.cpp
```diff
@@ -0,0 +1,306 @@
...
+//unit test for the test driver
+void driver_test()
+{
+	test_get_options();
+}
+
+void parse_cmd_line_and_compare(const std::string& cmd_line, const Options& expected,
+	const std::string& test_case)
+{
+	//this is meant to replicate the main() parameter char* argv[]
```
smurthys: The use of "this" needs a referent, without which we wouldn't know what the comment is about

```suggestion
	//args mimics parameter the argv parameter in main
```

test/driver-test/driver-test.cpp
```diff
@@ -0,0 +1,306 @@
...
+
+	Options received = get_options(args, size);
+	compare(received, expected, test_case);
+}
+
+
+//unit test for get_options
+void test_get_options()
+{
+	Options o;
```
smurthys: Rename `o` to something like `template_options` because that is what it is. The name `o` does not carry any meaning that is consistent with the persistent use of that variable.

test/driver-test/driver-test.cpp
```diff
@@ -0,0 +1,306 @@
...
+	o.command_name = "array_test";
+
+	using std::string;
+	
+	//default arguments
+	const string cmd_line_0("C:/Libraries/stl-lite/array_test.exe");
+	Options cmd_line_0_expected = o;
```
smurthys: There is no need to create a different Options instance for each cmd-line tested. Instead use a single instance and call it something like `expected_options` and reuse it throughout

test/driver-test/driver-test.cpp
```diff
@@ -0,0 +1,306 @@
...
+
+	const string cmd_line_2("C:/Libraries/stl-lite/array_test.exe -h yes");
+	Options cmd_line_2_expected = o;
+	cmd_line_2_expected.header = true;
+	cmd_line_2_expected.header_text = "Running $suite";
+	parse_cmd_line_and_compare(cmd_line_2, cmd_line_2_expected, "cmd_line_2");
+}
+
+//compares received Options object data members to expected Options object data members
+//see fn declaration of is_true for the behavior of the comparison
```
smurthys: Simplify comment; no need to comment about is_true

```suggestion
//perform pair-wise comparison of members in received options and expected options
```

test/driver-test/driver-test.cpp
```diff
@@ -0,0 +1,306 @@
...
+	test_get_options();
+}
+
+void parse_cmd_line_and_compare(const std::string& cmd_line, const Options& expected,
+	const std::string& test_case)
+{
+	//this is meant to replicate the main() parameter char* argv[]
+	//size 100 was chosen to ensure enough room for many options in a single command-line.
+	char* args[100];
+	std::vector<std::string> v = split(cmd_line, ' ');
```
smurthys, line 76: I believe a vector of `string_view` is better. I'm not sure, but with that change we might run into `const`ness of `data()`, but that can be addressed, and is worth doing so over creating a vector of strings

smurthys:
@Ndemco You probably have noticed that I have left some comments _in situ_. Overall, I think the revisions make the unit tests pretty straightforward. I suggest reviewing my comments and replying or changing and push the changes. Also, at that time, please say what else needs to be done: It seems like more command lines need to be added to `test_get_options`, but I could be wrong about that.

test/driver-test/driver-test.cpp
```diff
@@ -0,0 +1,306 @@
...
+
+	using std::string;
+	
+	//default arguments
+	const string cmd_line_0("C:/Libraries/stl-lite/array_test.exe");
+	Options cmd_line_0_expected = o;
+	parse_cmd_line_and_compare(cmd_line_0, cmd_line_0_expected, "cmd_line_0");
+
+	//single name-value option supplied
+	const string cmd_line_1("C:/Libraries/stl-lite/array_test.exe -h no");
```
smurthys: Just as with expected options, there is no need to create a different string instance for each cmd-line. 

PS: There is an advantage to having different string and Options instance for each cmd-line if they are `constexpr`. but `std::string` is not `constexpr` in C++17.

test/driver-test/driver-test.cpp
cmcmone: I didn't notice this problem. Thanks @Ndemco  for your clarify.

test/driver-test/driver-test.cpp
Ndemco: This is what I did in my first version in this PR, but I found it difficult to keep track of the values of each option data member. For example, `header` may have been set to false 3 command lines ago and not changed, and the current command line requires `header` to be true. It is hard to dig back to find out what the value of `header` in `expected_options` currently is and makes the code very error-prone, especially if something needs to be changed or new command-lines need to be added down the line. That is why I implemented the `reset_default_options` function, so after every edit made to `expected_options` (in this case, expected options was called `o`) i can call `reset_default_options` so the value of `expected_options` is known before setting the next expected values.

test/driver-test/driver-test.cpp
Ndemco: Perhaps the best route is to make the expected options instances `constexpr`? What do you think?

test/driver-test/driver-test.cpp
Ndemco: I agree, `string_view` would be better to use.

test/driver-test/driver-test.cpp
smurthys: At the risk of being wrong, I worry you are overthinking this: set `template_options` to have values that are most frequent. At each step copy `template_options` to `expected_options` and change the value of only those fields in `expected_options` that have to be different from the template. You do this for each cmd-line to test. What am I missing?

test/driver-test/driver-test.cpp
smurthys: No, Options instance can't be `constexpr` because it has `std::string` data members.

test/driver-test/driver-test.cpp
Ndemco: You are right, I am overthinking it. I understand now. My mistake!

Ndemco:
@smurthys 

> It seems like more command lines need to be added to `test_get_options`, but I could be wrong about that.

You are correct, there is a long comment at the bottom of `driver-test.cpp` that describes all the test cases that still need to be added to `test_get_options` (these are the command lines that were in the now removed `generate_cmd_line_and_options_pair` function). I wanted to make sure we agreed on the behavior and format of `test_get_options` before implementing it to every test case.

After implementing every test case described in the comment, I will take a deeper dive into all the command lines we have and see if there are more that need to be added.

test/driver-test/driver-test.cpp
Ndemco: I think there is an extra "parameter" in your suggestion.

Did you mean "args mimics the argv parameter in main"?

I will push that change but feel free to add another suggestion if you meant something else.

test/driver-test/driver-test.cpp
Ndemco: For now, the changes I've pushed do not use string_view as I'm still trying to reconcile the `const`ness of `string_view.data()` and the fact that `char* args[]` cannot be const since `get_options` requires a non-const array.

Ndemco:
I'm wondering if `test_get_options` and all functions specific to `test_get_options` should be in their own file to reduce clutter in `driver-test.cpp`.

cmcmone:
I have an idea about automatically generating command line, for example:

```c++
void get_header(std::string& cmd_line, Options& expected_options)
{
	bool header = static_cast<bool>(std::rand() % 2);
	if (header) {
		bool value = static_cast<bool>(std::rand() % 2);
		cmd_line.append(" -h ");
		cmd_line.append(value ? "yes" : "no");
		expected_options.header = value;
	}
}
```

- the advantage is that any number of command lines can be tested without manual input.
- the disadvantage is that it may make the test more complicated.

smurthys:
I made a few minor changes to driver-test. Please review.

- Revise file prologue
- Rename `test_get_options` to `test_get_options_nominal`
- Rename `parse_cmd_line_and_compare` to `test_cmd_line`
- Rename `compare` to `verify_options`
- Revise function prologue

smurthys:
Pending the use of std::string_view in parsing cmd-line, I'm OK with the code in this PR. However, I have not verified if the expected options are set up correctly for each cmd-line. I can spend some time tomorrow morning, but I think we all should carefully look at them.

@cmcmone I like your idea of generating cmd-line. The technique you propose is fairly well known and is the core of "fuzz" testing. The challenge in this case will be generating valid combinations, but it can be done. However, I recommend we consider that approach after this PR, at an appropriate time.

Ndemco:
@smurthys 
> I made a few minor changes to driver-test.

I like the function name changes. Thanks for that.

cmcmone:
Thanks @Ndemco @smurthys for the contribution.

- I notice that [L14-l17](https://github.com/sigcpp/stl-lite/pull/27/files#diff-a652f65fb5216681fe6cd07d6e6d05baR14-R17) include `iostream` and `string_view`, but I did see any else you use them.

- include `<cstdint>` because use `USHRT_MAX` in [L245](https://github.com/sigcpp/stl-lite/pull/27/files#diff-a652f65fb5216681fe6cd07d6e6d05baR245)

test/driver-test/driver-test.cpp
```diff
@@ -0,0 +1,298 @@
...
+	cmd_line = "C:/Libraries/stl-lite/array_test.exe -h yes";
+	expected_options = template_options;
+	expected_options.header = true;
+	expected_options.header_text = "Running $suite";
+	test_cmd_line(cmd_line, expected_options, "cmd-line 2");
+
+	cmd_line = "C:/Libraries/stl-lite/array_test.exe -p none";
+	expected_options = template_options;
+	expected_options.prm = pass_report_mode::none;
+	test_cmd_line(cmd_line, expected_options, "cmd_line_3");
```
cmcmone:
```suggestion
	test_cmd_line(cmd_line, expected_options, "cmd_line 3");
```

smurthys:
> I'm wondering if `test_get_options` and all functions specific to `test_get_options` should be in their own file to reduce clutter in `driver-test.cpp`.

I don't see any clutter @Ndemco. Could you point to some examples?

smurthys:
I just noticed that the PR was attempting to merge into `dev`, which is incorrect. I edited the PR so that the base is `driver-test` as it should be.

**Important:** please check the base at PR at creation. GitHub defaults the base to the default branch, which is `dev` in this repo.

test/driver-test/driver-test.cpp
cmcmone: > I believe a vector of `string_view` is better. I'm not sure, but with that change we might run into `const`ness of `data()`, but that can be addressed, and is worth doing so over creating a vector of strings

I am curious, how do we cast `const char*` to `char*`?

Ndemco:
> I just noticed that the PR was attempting to merge into dev

Sorry about that. I asked you on Teams about which branch I should be merging into and you sent me a screenshot of `get-options-nominal` merging into `dev` so I assumed that was what you wanted.

Ndemco:
@smurthys 
> I don't see any clutter @Ndemco. Could you point to some examples?

It is not very cluttered now but I imagine it will get cluttered as we add more to `driver_test`.

cmcmone:
> It is not very cluttered now but I imagine it will get cluttered as we add more to `driver_test`.

:smiley: I think `test_get_options_nominal` is not cluttered, but it's just too long. But if we split it, it feels like it breaks the integrity. 

[F.3: Keep functions short and simple](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f3-keep-functions-short-and-simple)

smurthys:
> Thanks @Ndemco @smurthys for the contribution.
> 
> * I notice that [L14-l17](https://github.com/sigcpp/stl-lite/pull/27/files#diff-a652f65fb5216681fe6cd07d6e6d05baR14-R17) include `iostream` and `string_view`, but I did see any else you use them.
> * include `<cstdint>` because use `USHRT_MAX` in [L245](https://github.com/sigcpp/stl-lite/pull/27/files#diff-a652f65fb5216681fe6cd07d6e6d05baR245)

Good catch @cmcmone. BTW, `USHRT_MAX` is declared in `<climits>`. BTW, [L245](https://github.com/sigcpp/stl-lite/pull/27/files#diff-a652f65fb5216681fe6cd07d6e6d05baR245) presents an interesting challenge for this PR because the data type of fail threshold has changed. I am right now thinking of a way to address this issue.

EDIT: I reviewed possible changes to avoid having to use `USHRT_MAX` here, but decided that any change causes conflicts either now or later. I think we can address this issue slightly differently later.

smurthys:
> > I just noticed that the PR was attempting to merge into dev
> 
> Sorry about that. I asked you on Teams about which branch I should be merging into and you sent me a screenshot of `get-options-nominal` merging into `dev` so I assumed that was what you wanted.

My bad then. I was only answering the question of which way the directions go. I assumed you were aware of the branch pattern. I recall discussing that `driver-test` branch encompasses all development related to testing the driver. No issues though.

smurthys:
> > It is not very cluttered now but I imagine it will get cluttered as we add more to `driver_test`.
> 
> 😃 I think `test_get_options_nominal` is not cluttered, but it's just too long. But if we split it, it feels like it breaks the integrity.
> 
> [F.3: Keep functions short and simple](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f3-keep-functions-short-and-simple)

I believe I have already mentioned a couple of time in this PR that it is perfectly fine to partition `test_get_options_nominal` to smaller functions. I don't think that the partitioning breaks the function's "integrity".

smurthys:
I assume you have noticed the alert that this branch has conflicts that must be resolved.

image

test/driver-test/driver-test.cpp
```diff
@@ -0,0 +1,298 @@
...
+	expected_options.summary = true;
+	expected_options.prm = pass_report_mode::none;
+	expected_options.fail_threshold = 2;
+	expected_options.fom = file_open_mode::new_file;
+	expected_options.output_filepath = "a_file.out";
+	expected_options.suites_to_run = "tester";
+	test_cmd_line(cmd_line, expected_options, "cmd-line 26");
+
+	//edge case: fail threshold
+	cmd_line = "C:/stl-lite/array_test.exe -t 65535"; //USHRT_MAX: compiler specific
```
smurthys: Use the macro I snuck into `get_fail_threshold` in `options.cpp`. It exists precisely to deal with data-type related issues.

```suggestion
	cmd_line = "C:/stl-lite/array_test.exe -t max"
```

Ndemco: Thanks for the suggestion @smurthys, I had forgotten about that macro.

test/driver-test/driver-test.cpp
Ndemco: Good catch @cmcmone. It should actually be `cmd-line 3`. I must have missed that one when I was changing them, thanks for pointing that out.

cmcmone:
> Good catch @cmcmone. BTW, `USHRT_MAX` is declared in `<climits>`. BTW, [L245](https://github.com/sigcpp/stl-lite/pull/27/files#diff-a652f65fb5216681fe6cd07d6e6d05baR245) presents an interesting challenge for this PR because the data type of fail threshold has changed. I am right now thinking of a way to address this issue.

Thanks @smurthys pointing out this error. I didn't realize that <climits> is a part of the [C numeric limits interface.](https://en.cppreference.com/w/cpp/types/climits)

> EDIT: I reviewed possible changes to avoid having to use `USHRT_MAX` here, but decided that any change causes conflicts either now or later. I think we can address this issue slightly differently later.

For this reason, I think @Ndemco can create an issue to record this, so we won't forget it in the later version.

cmcmone:
> > > It is not very cluttered now but I imagine it will get cluttered as we add more to `driver_test`.
> > 
> > 
> > 😃 I think `test_get_options_nominal` is not cluttered, but it's just too long. But if we split it, it feels like it breaks the integrity.
> > [F.3: Keep functions short and simple](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f3-keep-functions-short-and-simple)
> 
> I believe I have already mentioned a couple of time in this PR that it is perfectly fine to partition `test_get_options_nominal` to smaller functions. I don't think that the partitioning breaks the function's "integrity".

Sorry for not noticing the previous comment, I re-checked this PR comment again.
OK, if so, then we can split this function.

cmcmone:
> I assume you have noticed the alert that this branch has conflicts that must be resolved.
> 
> image

I resolved the conflict(first time), but I’m not sure if I did it correctly. Please let me know if there has a problem, then I will undo this merge.

cmcmone:
I think I can make some contribution to this PR, so I made the following changes:
- Change cmd_line_3 to cmd_line 3  14a01fa
- Fixed missing a semicolon  485c13b
- Remove useless header files (`iostream` and `string_view`)  4bb24ad
- [Fixed conflicts](https://github.com/sigcpp/stl-lite/pull/27/#issuecomment-646920680)  f333310
- [Split test function](https://github.com/sigcpp/stl-lite/pull/27/#issuecomment-646919622)  a02a142

Please let me know if you have any questions.  I can undo these mergers at any time.

I have another suggestion is that we don’t need the third parameter of `test_cmd_line`, we can generate this string in the `test_cmd_line` function and pass it to `verify_options`.

Ndemco:
@smurthys 
> I reviewed possible changes to avoid having to use USHRT_MAX here, but decided that any change causes conflicts either now or later. I think we can address this issue slightly differently later.

Thanks for investigating the issue, I agree this is something we can address later. I like @cmcmone's recommendation of creating an issue. If you agree, I can create an issue on this to be resolved later.

Ndemco:
@cmcmone 
> For this reason, I think @Ndemco can create an issue to record this, so we won't forget it in the later version.

I agree creating an issue is the best way to go about this. Following @smurthys approval, I will create an issue for this.

Ndemco:
Thanks @cmcmone for all the changes you made. I think your changes improved the organization of `driver-test.cpp` significantly. I made one small change which was changing `cmd_line 3` to `cmd-line 3`.

I am interested to see how you would generate the `test_case` string in the `test_cmd_line` function.

cmcmone:
> Thanks @cmcmone for all the changes you made. I think your changes improved the organization of `driver-test.cpp` significantly. I made one small change which was changing `cmd_line 3` to `cmd-line 3`.

:sweat_smile: I habitually write the underline. Thanks @Ndemco

> I am interested to see how you would generate the `test_case` string in the `test_cmd_line` function.

I update the `test_cmd_line` function and generate the `test_case` string in it.   80bf78f

smurthys:
> @cmcmone
> 
> > For this reason, I think @Ndemco can create an issue to record this, so we won't forget it in the later version.
> 
> I agree creating an issue is the best way to go about this. Following @smurthys approval, I will create an issue for this.

 believe merging dev into this branch addresses this issue. After merging, `USHRT_MAX` would need to be replaced with `INT_MAX` in `driver_test`. Unrelated to this, I think merging dev into this branch does cause an expected conflict, one that is easily resolved.

smurthys:
So, I looked at splitting cmd-line into string_view instead of string. The process revealed some unrelated issues in `get_options`, which I logged as Issue #29.

Back to splitting cmd-line into string_view in `test_cmd_line`: Though the change needed to use string_view is pretty straightforward including addressing const-ness, splitting a single long cmd-line string requires spaces to be replaced with the null character otherwise, the elements of the args array become parts of one long C-string. The situation is best demonstrated in a call. You could see it for yourself, but I suggest not spending that effort at this point. We should rather continue to split using string and close this PR first.

smurthys:
> I think I can make some contribution to this PR, so I made the following changes:
> 
> * Change cmd_line_3 to cmd_line 3  [14a01fa](https://github.com/sigcpp/stl-lite/commit/14a01faeb6d189293ab4459746fe8b589a2e395f)
> * Fixed missing a semicolon  [485c13b](https://github.com/sigcpp/stl-lite/commit/485c13bf8242d8f53b77bcfe0f06d8d3359074d7)
> * Remove useless header files (`iostream` and `string_view`)  [4bb24ad](https://github.com/sigcpp/stl-lite/commit/4bb24ad11c3f6c775ab9696938a3b9d9036b2910)
> * [Fixed conflicts](https://github.com/sigcpp/stl-lite/pull/27/#issuecomment-646920680)  [f333310](https://github.com/sigcpp/stl-lite/commit/f333310174d6b01cb85c2635867e3beab2ecb50c)
> * [Split test function](https://github.com/sigcpp/stl-lite/pull/27/#issuecomment-646919622)  [a02a142](https://github.com/sigcpp/stl-lite/commit/a02a14252cbec790a352674f95a878892227c198)
> 
> Please let me know if you have any questions. I can undo these mergers at any time.
> 
> I have another suggestion is that we don’t need the third parameter of `test_cmd_line`, we can generate this string in the `test_cmd_line` function and pass it to `verify_options`.

Thanks @cmcmone for jumping in. The changes look good. IMO it would have been better to [study fuzz testing](https://github.com/sigcpp/stl-lite/pull/27#issuecomment-646344065) before creating the smaller test functions, but the current code is acceptable. 

BTW, in commit in commit f333310, I don't see changes relevant to the merge [conflicts identified](https://github.com/sigcpp/stl-lite/pull/27#issuecomment-646859102). The conflicts were in VS project and project filters files, but the commit shows changes only in `driver-test.cpp`. Am I seeing this wrong?

smurthys:
Thanks @Ndemco for creating the tests. They provide a fair coverage of the nominal behavior of `get_options`. Thanks @cmcmone for breaking down functions and making other edits.

I just pushed two changes: 
- Made template options a file-scoped static constant, and removed `template_options` parameter from test functions.
- Renamed function `test_get_options_combine_random` to `test_get_options_combine_arbitrary`. The commit message has the rationale

I have the following summary thoughts about the  the test driver:
- The functions appear to be placed in some undecipherable order: It is common practice to list related functions together; and to place first the function that is used first. This is not the case currently. Also, using the common practice, I recall moving the function `test_cmd_line` to the bottom in b593d08, but it seems that change was reverted. Not a big deal, but we need to make and practice a convention. The common practice is as good as any.
- While the current tests provide good coverage, they do not test if the `command_name` member is retrieved correctly, because all tests use the same program file path. I recommend adding a couple of tests just to test that member.

I have not been able to verify the quality of all the tests, but eyeballing, they seem good and correctly set up.

Other than the two thoughts I share above, I feel these changes look good. I recommend merging `dev` into this branch which I see will cause three simple and obvious conflicts, and all are fixed rather easily: just make sure there are three tests suites in suites.cpp after merging, and the project has a filter for each test suite under "Source Files".

cmcmone:
> > @cmcmone
> > > For this reason, I think @Ndemco can create an issue to record this, so we won't forget it in the later version.
> > 
> > 
> > I agree creating an issue is the best way to go about this. Following @smurthys approval, I will create an issue for this.
> 
> believe merging dev into this branch addresses this issue. After merging, `USHRT_MAX` would need to be replaced with `INT_MAX` in `driver_test`. Unrelated to this, I think merging dev into this branch does cause an expected conflict, one that is easily resolved.

Yes, I think we can merge current branch to `driver-test` first, and then merge dev to `driver-test`.

cmcmone:
> Back to splitting cmd-line into string_view in `test_cmd_line`: Though the change needed to use string_view is pretty straightforward including addressing const-ness, splitting a single long cmd-line string requires spaces to be replaced with the null character otherwise, the elements of the args array become parts of one long C-string. The situation is best demonstrated in a call. You could see it for yourself, but I suggest not spending that effort at this point. We should rather continue to split using string and close this PR first.

Thanks @smurthys for explaining, but I'm not sure I fully understand, maybe we can make it clear the next meeting.

cmcmone:
> BTW, in commit in commit [f333310](https://github.com/sigcpp/stl-lite/commit/f333310174d6b01cb85c2635867e3beab2ecb50c), I don't see changes relevant to the merge [conflicts identified](https://github.com/sigcpp/stl-lite/pull/27#issuecomment-646859102). The conflicts were in VS project and project filters files, but the commit shows changes only in `driver-test.cpp`. Am I seeing this wrong?

I’m sure I modified the VS project and project filters files, and after I submit it, it only shows this  [Merge branch 'driver-test' into get-options-nominal](https://github.com/sigcpp/stl-lite/pull/27/commits/f333310174d6b01cb85c2635867e3beab2ecb50c), nothing else. I don't know why.

cmcmone:
> I just pushed two changes:
> 
> * Made template options a file-scoped static constant, and removed `template_options` parameter from test functions.
> * Renamed function `test_get_options_combine_random` to `test_get_options_combine_arbitrary`. The commit message has the rationale

Thanks @smurthys for that change, I agree almost, just curious that the `get_template_options` function also needs to be set to static?

cmcmone:
> * The functions appear to be placed in some undecipherable order: It is common practice to list related functions together; and to place first the function that is used first. This is not the case currently. Also, using the common practice, I recall moving the function `test_cmd_line` to the bottom in [b593d08](https://github.com/sigcpp/stl-lite/commit/b593d081864b658dd032924e43933897d2924a73), but it seems that change was reverted. Not a big deal, but we need to make and practice a convention. The common practice is as good as any.

Thanks @smurthys to pointing out, I think it's because I messed up the order when I split the function. Should I change it? It might be inappropriate if I make changes before ready merge.

cmcmone:
> * While the current tests provide good coverage, they do not test if the `command_name` member is retrieved correctly, because all tests use the same program file path. I recommend adding a couple of tests just to test that member.

Is it possible to create an array to save `command_name`, then test it?

smurthys:
@cmcmone 
> just curious that the get_template_options function also needs to be set to static?

A function that should be callable only within the translation unit (TU) should be marked static, just like a file-scope variable. So, technically the many of the `test_` functions should also be marked `static` but I'd leave them the way they are, for now.

smurthys:
@cmcmone 

> I think we can merge current branch to driver-test first, and then merge dev to driver-test

No. I'm saying merge `dev` into this branch **before** merging this PR (no matter to which branch this PR merges). I suggest we do this during Monday's call so we can go over merge conflicts, and @Ndemco should merge `dev` into this branch. We need to make sure that all changes are complete before Monday's call.

smurthys:
> > * While the current tests provide good coverage, they do not test if the `command_name` member is retrieved correctly, because all tests use the same program file path. I recommend adding a couple of tests just to test that member.
> 
> Is it possible to create an array to save `command_name`, then test it?

I' afraid I don't follow. Do you mean create an array with many program file paths and use the array elements to test if command_name returned is correct? If so, yes. An array with program file paths and a corresponding array with expected command names can be used. Alternatively, a map with program file path as value and command_name as key can also be used.

However, the choice depends on how many program file paths are being checked. We should always keep the solution simple and use the simplest means. For example, if we test only a couple of program file paths, don't use arrays or maps. Just directly set the values and test. That is, start with designing the tests first and worry about implementing it later.

smurthys:
> > * The functions appear to be placed in some undecipherable order: It is common practice to list related functions together; and to place first the function that is used first. This is not the case currently. Also, using the common practice, I recall moving the function `test_cmd_line` to the bottom in [b593d08](https://github.com/sigcpp/stl-lite/commit/b593d081864b658dd032924e43933897d2924a73), but it seems that change was reverted. Not a big deal, but we need to make and practice a convention. The common practice is as good as any.
> 
> Thanks @smurthys to pointing out, I think it's because I messed up the order when I split the function. Should I change it? It might be inappropriate if I make changes before ready merge.

At this point, I feel we shouldn't bother changing anything that is not necessary. Going forward, we should just remember to do it right the first time.

cmcmone:
> @cmcmone
> 
> > just curious that the get_template_options function also needs to be set to static?
> 
> A function that should be callable only within the translation unit (TU) should be marked static, just like a file-scope variable. So, technically the many of the `test_` functions should also be marked `static` but I'd leave them the way they are, for now.

Thanks @smurthys for the explain.

So the function is defined as static is like "hidden", prevent external files from accessing it?

cmcmone:
> > > * While the current tests provide good coverage, they do not test if the `command_name` member is retrieved correctly, because all tests use the same program file path. I recommend adding a couple of tests just to test that member.
> > 
> > 
> > Is it possible to create an array to save `command_name`, then test it?
> 
> I' afraid I don't follow. Do you mean create an array with many program file paths and use the array elements to test if command_name returned is correct? If so, yes. An array with program file paths and a corresponding array with expected command names can be used. Alternatively, a map with program file path as value and command_name as key can also be used.
> 
> However, the choice depends on how many program file paths are being checked. We should always keep the solution simple and use the simplest means. For example, if we test only a couple of program file paths, don't use arrays or maps. Just directly set the values and test. That is, start with designing the tests first and worry about implementing it later.

Make sense, I also agree with your opinion

>That is, start with designing the tests first and worry about implementing it later.

thank you for your reply.

Ndemco:
Thanks @smurthys and @cmcmone for the contributions you made yesterday. I've read through the new suggestions and hope to address all of the pending issues and discussions in this comment.

* While @cmcmone attempted to resolve the merge conflicts, there still seemed to be merge conflicts that were present. The last comment I see about it is this:
> I’m sure I modified the VS project and project filters files, and after I submit it, it only shows this Merge branch 'driver-test' into get-options-nominal, nothing else. I don't know why.

However, I'm assuming this was resolved as there doesn't seem to be any conflicts present anymore.

* Thanks @smurthys for figuring out how to use `string_view`s, I agree it is probably best to go over the required changes in a call and merge the branch using `string`s for now.

* I've looked through all the changed code, it is much more readable and organized than it was when this PR was created; thank you both for the improvements you've made. One issue I see that I think needs to be resolved before we merge as it is an important part of design: because the `test_case` parameter in `verify_options` is now automatically generated, it is very hard to know which `test case` correlates with which command line since it is not explicitly written in each call to `test_cmd_line`. This makes it difficult to find exactly which command-line failed when running tests. I believe this issue needs to be resolved before merging this branch as the ability to easily find which test case failed is an important part of design. I can revert some of the changes and add the `test_case` parameter back to `test_cmd_line` so the command-line number can be explicitly written, but I know @cmcmone  put a lot of work into the code the way it is now and I would not want to change or revert anything without first hearing your opinions. Perhaps I am not seeing it correctly or perhaps there is a better way to solve this.

* I pushed a change that adds the function `test_get_options_command_line` which tests a few different file paths and file names. I also changed the file path/name in two test cases in `test_get_options_combine_arbitrary` and added test cases in `test_get_options_edge_case` to test for a very very long `command_name`. Thanks @smurthys for catching this.

I think this addresses most of the ongoing conversations in this PR and my own thoughts. If I missed something and you would like my input, please let me know. Great work everyone!

cmcmone:
>because the test_case parameter in verify_options is now automatically generated, it is very hard to know which test case correlates with which command line since it is not explicitly written in each call to test_cmd_line. 

Thanks to @Ndemco for pointing out this serious bug. I didn't realize this before modifying the code. I am sorry for my mistake.

smurthys:
@Ndemco Good changes. I suggest adding the following tests where the program file path:
- does not have a drive letter or path, say `driver.exe`
- does not have an extension , `multi-drive` (this is legal)
- has a relative path and a non-EXE extension, `bin/driver.com`

I agree that automating test identification results in information loss. I do recommend adding explicit case identification back. However, I like to remind again that the test structure will be different if [fuzz testing](https://github.com/sigcpp/stl-lite/pull/27#issuecomment-646344065) is adopted (which for some reason hasn't received any love in this thread 😿)

Ndemco:
@cmcmone 
> Thanks to @Ndemco for pointing out this serious bug. I didn't realize this before modifying the code. I am sorry for my mistake.

I did not realize this either until I had created a buggy command line and couldn't find which one had the bug. We both approved of this so it is both of our mistake!

Ndemco:
@smurthys I've implemented your suggestions into `test_get_options_command_name` thanks for pointing them out.

I've done a little research on fuzz testing but I don't have a great understanding of how it is implemented. I would love to learn more about it!

Ndemco:
I've just made a push the makes the following changes:
* Add @smurthys's [suggested test cases](https://github.com/sigcpp/stl-lite/pull/27#issuecomment-647138486) to `test_get_options_command_name`.
* Revert the code back to explicit case identification
* Changed the order some of the tester functions are defined so `cmd-line 1` - `cmd-line 37` are defined in order of their number; this makes it easier to search for the specific command line that triggered the error.

WIth these changes, I think this PR will be ready to merge after our meeting today. Thanks @smurthys and @cmcmone for all the hard work!

smurthys:
Thanks @Ndemco for making the changes. The changes look good. I look forward to today's call.

cmcmone:
Thanks @Ndemco for those changes.

smurthys:
Re-opened because I think @cmcmone accidentally closed the PR before it was approved and merged.

Ndemco:
I just tested all the changes and everything seems to work as intended.

cmcmone:
> Re-opened because I think @cmcmone accidentally closed the PR before it was approved and merged.

Yes, I accidentally closed it, sorry about it. @smurthys

smurthys:
Thanks @Ndemco. I too tested and all three test suites run successfully. Excellent job everyone.

calebGarrick:
I also successfully pulled and tested the test suites. Great work!

cmcmone:
Thanks @Ndemco for merge, everything looks good, but I think there have some tiny issue need to fix

- [L410](https://github.com/sigcpp/stl-lite/pull/27/files#diff-a652f65fb5216681fe6cd07d6e6d05baR410) since we use `INT_MAX`, so I think we must include `<climits>`
- Format the code, [L42](https://github.com/sigcpp/stl-lite/pull/27/files#diff-a652f65fb5216681fe6cd07d6e6d05baR42) and [L150 ](https://github.com/sigcpp/stl-lite/pull/27/files#diff-a652f65fb5216681fe6cd07d6e6d05baR215) need to add an empty line(two empty lines between functions)

Ndemco:
> 
    L410 since we use INT_MAX, so I think we must include <climits>
    Format the code, L42 and L150 need to add an empty line(two empty lines between functions)


Good catch @cmcmone, I just pushed changes that address those issues. I did not find the issue on L150 that you mentioned, but I went through the file and made sure all function definitions had two empty lines after them.

